Issue #8, opened 2019-01-30
Investigate Containerization


MekWarrior:
Use deepcell-tf as roadmap to investigate containerization. It is likely a graphics port will need to be exposed.

MekWarrior:
Solved by #30

PR #30, opened 2019-08-01
Dockerize Caliban

icejannie:
Added the ability for Caliban to run in a Docker container.

To be able to view the graphical application, we needed an X server, a window manager, and a means to access the container's display. A bootstrap script was written to start Xvfb in the background, add a fluxbox window manager, and access the display using a VNC server. 

Dockerfile uses ubuntu:16.04 as base image and installs necessary packages before calling the bootstrap script. 

README.md has been updated to include Docker instructions and new Caliban functions.


icejannie:
Also would like to request to have the repository renamed to 'caliban'. Docker only allows lowercase letters for image names.

MekWarrior:
Solves Issue #8

Issue #27, opened 2019-07-27
Fill holes in annotation mask


geneva-miller:
Poor annotations can have holes in the middle of a cell. Fixing these does not require updating lineage information, just updating the annotation. skimage.morphology has some functions that may be helpful here: https://scikit-image.org/docs/dev/api/skimage.morphology.html (notably, fill_small_holes and flood_fill).

MekWarrior:
Solved by #31

Issue #20, opened 2019-07-10
Add new mask to annotation


geneva-miller:
Useful in cases where annotation is incorrect (missing pixels) or user has erroneously deleted a cell mask in a frame.

This option should:

- require user input to determine where annotation should be created
    - two clicks to determine corners of bounding box that contains cell to be annotated
    - third click to determine seed location for watershed transform
- create new mask in that frame using largest cell value in movie + 1 (new unique mask)
- create new lineage information corresponding to cell mask

MekWarrior:
Solved by #31

Issue #16, opened 2019-03-13
Caliban should never make an empty track


MekWarrior:
Before saving, the program should check to see if there are any tracks that do not contain any frames and delete them.

geneva-miller:
Sort of fixed with #28 . Empty tracks are removed upon save, which is useful for fixing otherwise correct trk files with bad lineage info, and for preventing empty tracks from being saved into future lineages. Lineage-editing functions should be reviewed to fix this behavior instead of relying on catching it at save.

MekWarrior:
Solved by #28 and #31

Issue #6, opened 2019-01-10
Correct for single pixel segmentation errors


MekWarrior:
Allow for label deletion in a frame (in case of 1 pixel segmentation). Should also allow for expansion of labels (in case of missing pixels)

geneva-miller:
Label deletion added in #22 , still need way to add labels or expand labels for trk files

MekWarrior:
Solved by #31 - should port to .trk side

PR #31, opened 2019-08-01
Feature/zstack editor mode

geneva-miller:
Add npz editor to Caliban, with features focused on zstack curation. New features within npz editor include zstack prediction, hole-filling, annotation editing mode that allows for pixel-level segmentation changes with a brush tool, ability to cycle through features and channels.


MekWarrior:
Solves #27, #20, #16, #6

PR #72, opened 2019-12-03
Update browser trk mode

geneva-miller:
Fixes trk bugs analogous to npz bugs that were fixed in #63. Cleans up some html/css for both trk and npz files and reduces payload that application.py sends by removing edit_image, since trk mode no longer needs this.

### DIFF
--- a/browser/caliban.py
+++ b/browser/caliban.py
@@ -571,10 +571,6 @@ class TrackReview:
 
         self.scale_factor = 2
 
-        self.highlight = False
-        self.highlight_cell_one = -1
-        self.highlight_cell_two = -1
-
         self.color_map = plt.get_cmap('cubehelix')
         self.color_map.set_bad('red')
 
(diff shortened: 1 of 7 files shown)


browser/caliban.py
```diff
@@ -621,44 +594,27 @@ def readable_tracks(self):
...
-                    frame = np.ma.masked_equal(frame, self.highlight_cell_one)
-                if (self.highlight_cell_two != -1):
-                    frame = np.ma.masked_equal(frame, self.highlight_cell_two)
-
+            frame = self.tracked[frame][:,:,0]
+            frame = np.ma.masked_equal(frame, 0)
             return pngify(imgarr=frame,
                          vmin=0,
-                         vmax=self.num_tracks,
+                         vmax=max(self.tracks),
```
MekWarrior: this could be problematic if a large label is chosen. Ideally i think we get the length of unique ids.

MekWarrior: Moved investigation to issue  #73

PR #75, opened 2019-12-04
Consolidate browser actions to use just one POST request

geneva-miller:
browser/caliban.py keeps track of whether actions change the annotations or info, which application.py uses to send the appropriate payload (if frames change, send the updated frame, if info changes, send the updated info) in response to the action.

Also cleans up related code (removes unnecessary ajax/flask functions, cleans up TrackReview and ZStackReview attributes, use np.where for all pixel-editing actions).

### DIFF
--- a/browser/caliban.py
+++ b/browser/caliban.py
@@ -184,8 +184,6 @@ class ZStackReview:
         if in_original and not in_modified:
             self.del_cell_info(feature = self.feature, del_label = old_label, frame = frame)
 
-
-
     def action_trim_pixels(self, label, frame, x_location, y_location):
         '''
         get rid of any stray pixels of selected label; pixels of value label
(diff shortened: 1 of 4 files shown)


browser/caliban.py
```diff
@@ -205,12 +195,17 @@ def action_trim_pixels(self, label, frame, x_location, y_location):
...
+        comparison = np.where(img_trimmed != img_ann)
+        self.frames_changed = np.any(comparison)
+        #this action should never change the cell info
+
+        self.annotated[frame,:,:,self.feature] = img_trimmed
 
     def action_handle_draw(self, trace, edit_value, brush_size, erase, frame):
 
-        annotated = self.annotated[frame,:,:,self.feature]
+        annotated = np.copy(self.annotated[frame,:,:,self.feature])
```
MekWarrior: Verify this copy call behaves as expected. ie: shallow copy vs deep copy

MekWarrior: same for line 728

geneva-miller: np.copy works analogously to deep copy in this case (when annotated is modified, self.annotated is not modified)
https://docs.scipy.org/doc/numpy-1.16.0/reference/generated/numpy.copy.html

PR #105, opened 2020-03-13
Add elastic beanstalk-compatible version of browser application

geneva-miller:
Elastic beanstalk, which we use to deploy browser-caliban, has a few differences and additional configuration files required for it to run effectively. This set of changes allows both versions of the application to coexist instead of maintaining each on separate branches. Launching the app to elastic beanstalk remains straightforward as the included config files will properly configure the WSGI path such that elastic beanstalk uses the appropriate version of the Flask app.

Also applies Will's changes to application.py and other browser scripts, which had not been merged into master as they included elastic beanstalk-specific changes. This set of changes brings application.py up to date and uses python decouple for configuration/API key management.


browser/eb_application.py
```diff
@@ -0,0 +1,357 @@
...
+from flask import Flask, jsonify, render_template, request, redirect
+
+from helpers import is_trk_file, is_npz_file
+from caliban import TrackReview, ZStackReview
+import config
+
+# Create and configure the app
+application = Flask(__name__)  # pylint: disable=C0103
+app = application
+app.config.from_object("config")  # TODO: did this break by adding new env vars?
```
MekWarrior: Have we stress tested re: this todo?

geneva-miller: I am not sure exactly what that comment (Will's, I think) was referring to. Off the top of my head I think this step used to work by accessing environment variables from os, and the config.py was pretty different. Configuration no longer relies on any environment variables and is not broken so I think this is a non-issue.

Issue #57, opened 2019-11-05
Caliban should be able to open files with no annotations


geneva-miller:
It is possible that we will deal with npz (or trk?) files where no annotations exist. This could be because there are no objects that need to be annotated, or because objects that need annotations do not have corresponding annotations. Currently, the reshape npz (preprocessing) function in deepcell-toolbox addresses this by not saving npzs with empty annotation files. However, we may want to run "empty" annotation files through fig8 in the future, or even just check through files with caliban.

geneva-miller:
Addressed by #87 for desktop version, but still needs to be implemented for browser.

PR #111, opened 2020-03-14
Browser/fix/open empty annotations

geneva-miller:
Allows browser caliban to open npzs that don't contain annotations.

Is this something we would ever need for a trk file? I don't anticipate we would ever have trk files start empty with no annotations/lineage so no fix for that case is included.

### DIFF
--- a/browser/caliban.py
+++ b/browser/caliban.py
@@ -76,6 +76,21 @@ class ZStackReview:
         self.frames_changed = False
         self.info_changed = False
 
+    def get_max_label(self):
+        '''
+        Helper function that returns the highest label in use in currently-viewed
+        feature. If feature is empty, returns 0 to prevent other functions from crashing.
+        (Replaces use of self.num_cells to keep track of this info, should
+        also help with code flexibility.)
+        '''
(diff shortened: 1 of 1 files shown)


geneva-miller:
Should close #57 .

browser/caliban.py
```diff
@@ -77,6 +76,21 @@ def __init__(self, filename, input_bucket, output_bucket, subfolders):
         self.frames_changed = False
         self.info_changed = False
 
+    def get_max_label(self):
+        '''
+        Helper function that returns the highest label in use in currently-viewed
+        feature. If feature is empty, returns 0 to prevent other functions from crashing.
+        (Replaces use of self.num_cells to keep track of this info, should
```
MekWarrior: lets move this note re: self.num_cells to a PR comment rather than keep it in the code base

MekWarrior:
This shouldn't be required for .trk files as the annotations should be created prior to establishing any lineages. This is a procedural requirement and should be included in the documentation as such.

geneva-miller:
Okay, that's what I was thinking. Where in the documentation should that go?

PR #116, opened 2020-03-16
Browser/feature/disable page scroll

geneva-miller:
Prevent scroll wheel and keys space, up, and down from scrolling page when working on trk or npz files. Makes it easier for annotators to use scroll and these keybinds for their intended functionality.

### DIFF
--- a/browser/static/js/main_zstack.js
+++ b/browser/static/js/main_zstack.js
@@ -1093,6 +1093,10 @@ function action(action, info, frame = current_frame) {
 }
 
 function start_caliban(filename) {
+  // disable scrolling from scrolling around on page (it should just control brightness)
+  document.addEventListener('wheel', function(event) {
+    event.preventDefault();
+  }, {passive: false});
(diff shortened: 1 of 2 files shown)


browser/static/js/main_zstack.js
```diff
@@ -1093,6 +1093,20 @@ function action(action, info, frame = current_frame) {
 }
 
 function start_caliban(filename) {
+  // disable scrolling from scrolling around on page (it should just control brightness)
```
MekWarrior: not just brightness right? also scales color (e.g. label scale)

geneva-miller: That feature (adjusting color range of labels) is not included in browser. It would require server calls to adjust and reapply the colormap to the label array, then send back a png. Between the modified viridis colormap that browser uses and highlighting, I don't think it's necessary to include. That's why scrolling only affects brightness of raw image

MekWarrior: 👍

PR #119, opened 2020-03-23
95th percentile scaling

### LINKED ISSUE #104: rgb mode rendering/image thresholding issue
I can't figure out exactly what is causing this, but the newest batch of IF data has a weird thresholding artifact when displayed in caliban. Below is what I see when plotting the contents of the NPZ:

image

And here is what I see in caliban: 
image

It looks like the values are being capped, which is what leads to the contour-like effect. This image has values up to 800, which is brighter than normal MIBI images. However, when I rescaled the npz by dividing by 200, I got the same effect. I've attached an example NPZ here so you can take a look. 
[Eliot_Point17_row_0_col_0.npz.zip](https://github.com/vanvalenlab/caliban/files/4316487/Eliot_Point17_row_0_col_0.npz.zip)

ngreenwald:
closes #104

### DIFF
--- a/desktop/caliban.py
+++ b/desktop/caliban.py
@@ -3825,7 +3825,11 @@ class RGBNpz(CalibanWindow):
 
     def update_adjusted_raw(self):
         # self.raw may not be in 8 bit (0, 255) format; rescale without modifying original
-        rescaled_raw = rescale_intensity(self.raw[:,:,0:6], in_range = 'image', out_range = 'uint8')
+
+        # get value at 5th and 95th percentile for min and max rescaling, respectively
+        percentiles = np.percentile(self.raw[self.raw > 0], [5, 95])
+        rescaled_raw = rescale_intensity(self.raw[:,:,0:6], in_range=(percentiles[0], percentiles[1]),
+                                         out_range='uint8')
         rescaled_raw = rescaled_raw.astype('uint8')
 
         for c, adjust in enumerate(self.adjustments):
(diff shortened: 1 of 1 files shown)


geneva-miller:
Looks like this will essentially clip very dim and very bright pixels, nice. Have you used this on other files with various data ranges to confirm that it displays those well?

ngreenwald:
No, good call, let me test on some super low int images as well to make
sure it doesn't do anything weird.

ngreenwald:
Looks good to go! Tested it on IMC and MIBI (low counts) and two different IF platforms (high counts)

desktop/caliban.py
```diff
@@ -3825,7 +3825,11 @@ def helper_update_composite(self):
 
     def update_adjusted_raw(self):
         # self.raw may not be in 8 bit (0, 255) format; rescale without modifying original
-        rescaled_raw = rescale_intensity(self.raw[:,:,0:6], in_range = 'image', out_range = 'uint8')
+
+        # get value at 5th and 95th percentile for min and max rescaling, respectively
+        percentiles = np.percentile(self.raw[self.raw > 0], [5, 95])
+        rescaled_raw = rescale_intensity(self.raw[:,:,0:6], in_range=(percentiles[0], percentiles[1]),
+                                         out_range='uint8')
         rescaled_raw = rescaled_raw.astype('uint8')
```
MekWarrior: out of interest, what was the design choice behind using uint8?

geneva-miller: The uint8 array gets drawn directly as an RGB image (no colormap gets applied). The underlying raw data type doesn't get changed.

Issue #110, opened 2020-03-13
Browser improvements to make annotation easier


geneva-miller:
Collection of small changes that would improve ease of use and display:

- [x] disable effect of scroll on window: as with desktop, scrolling on the image changes the brightness. When viewing browser caliban directly, everything fits within the displayed page so this doesn't scroll across the page. When annotators access Caliban via iframe, however, using the scroll functionality or up/down keys (to adjust brush size) moves the page, which is annoying and may reduce the use of these tools. The elastic beanstalk app that is currently deployed has a fix implemented but this fix needs to be added to a branch that will get merged into master. #88 

- [ ] opacity of annotations: browser displays label/raw overlay by drawing label image on top of raw image with a fixed alpha value (easy to do in js). A side effect is that the black background also gets overlaid which can artificially dim the raw image, making it harder to judge details. We should investigate adding an alpha channel to the png data and/or changing the overlay method to produce clearer images. This may take some iteration to ensure that all images get displayed clearly.

- [x] add padding to canvas: mouse events are bound to canvas which makes drawing or thresholding on edge difficult. We should increase the size of the canvas by a fixed amount on all sides (similar to desktop caliban's padding around image) to address this. Functions that use canvas coordinates will also likely need to be updated to reflect this. Adding canvas padding may also be a useful precursor to zoom, as it is in desktop. #83 

- [x] split brush into its own file: brush objects should be imported from a common file to reduce code redundancy between slightly different caliban use cases

- [x] update brush display: browser brush should be adapted to be more similar to desktop brush (as it is in zoom feature branch). Ie, brush interior is translucent white and has solid white or red edge (depending on status of eraser). Visual cue of brush outline has been very helpful for annotating in desktop. Brush should still leave outline of where it has gone; this may or may not have edging around it as in desktop. See also #52 

- [x] update highlighting (related to brush visual update): currently highlighting recolors a label to red in both label- and pixel-editing modes. I think it's worth considering applying edge highlighting in pixel-editing mode instead, as with desktop. Likely, we would send an array containing edge information along with other image information upon updating frames. Applying the outlines to the image would occur in a similar manner as highlighting is currently applied in js. This one is a little less urgent since a way to highlight in pixel-editing mode does already exist

[...]

PR #121, opened 2020-03-24
Update browser contrast adjustment

geneva-miller:
Adds brightness adjustment (shift+scroll), which helps visualize phase images. Stores brightness/contrast adjustments separately per channel for npz files. 0 (zero) key resets both brightness and contrast values. Scroll direction, not magnitude, determines adjustment so behavior is consistent across trackpad/mouse use.


geneva-miller:
Addresses part of #110

Issue #109, opened 2020-03-13
Save user settings in MySQL to better profile successful jobs.


willgraf:
It seems like our users are having a hard time with some of the jobs. While I think we have a decent guess of why these jobs are taking too long (based on our own user experience), it might be good to save any user settings for each job, along with the time it took to complete the job (#71).

I think this could be helpful in looking into any common issues users have while working on these jobs.

geneva-miller:
Settings would need to be stored in a different table (?) since uploading a file also causes the corresponding object to be deleted from the database.

geneva-miller:
Having just the time taken per file would be a helpful first step for accurately pricing jobs, but I am unfamiliar with SQL and unsure about how to implement this.

willgraf:
We can create a couple of columns, `createdAt` and `updatedAt`, to track the timestamps. A nice thread about it [here](https://stackoverflow.com/questions/3488536/mysql-create-time-and-update-time-timestamp).

geneva-miller:
To clarify, what I am uncertain about is how we should add another table to what we currently have (or whatever approach you think would work better). Is this something you could take a stab at, or could walk me through? Implementing the SQL code to do so is not intuitive to me.

geneva-miller:
I've pushed a branch (browser/feature/timestamp_tracking) with a first pass attempt at implementing this. @willgraf I'd appreciate it if you could take a look at it before I try running it.

geneva-miller:
A preliminary version of this is now deployed, which tracks timestamps for opening, updating, and submitting the file. @MekWarrior @willgraf  What other metrics do y'all think could be helpful here? Ones I am thinking of include:
-longest time between updates (someone steps away from file in the middle of it?)
-number of updates (increment each time the file is updated, as a proxy for how much "work" each file takes)
-don't update "updatedAt" timestamp when file is submitted (will show us time between last action and file submission)
-time between file creation and first update (proxy for time spent reading instructions)

willgraf:
To implement some of the above will take a bit of extra effort, as the `updatedAt` field I created by default is automatically updated every time there is a change.  It will not be as simple to update it occasionally.

That said, we could add `firstUpdate` and `lastUpdate` timestamps that we update manually at the given times.

Otherwise, maybe consider:
- objects per image
- total updates
- count of mistakes? like undos or erases or something?
- average brush size
- ....?

geneva-miller:
I think the metrics from #124 should be sufficient, but we can return to this later if needed.

PR #124, opened 2020-03-25
Initial timestamp tracking

geneva-miller:
Adds some tracking metrics, including timestamps, to SQL database used with browser caliban. Instead of deleting row from database when file is finished, replaces long data blob with null to clear space. Keeps track of timestamps such as when file is opened and submitted. Keeps track of number of updates submitted to file. Helpful for determining how long files take to annotate, and how that time breaks down. Should also help database maintenance by making it easier to drop stale rows.

I think this is a solid start re: #109. I would be satisfied with closing that issue and opening a separate issue later if we need more advanced metrics or other changes to the SQL code.

### DIFF
--- a/browser/eb_application.py
+++ b/browser/eb_application.py
@@ -364,6 +364,12 @@ def delete_project(conn, project_id):
     '''
     cur = conn.cursor()
 
+    sql = ''' UPDATE {tn}
+              SET lastUpdate = updatedAt
+              WHERE id = {my_id}
+          '''.format(tn = 'projects', my_id = project_id)
+    cur.execute(sql)
+
     timesql = ''' UPDATE {tn}
                   SET finished = CURRENT_TIMESTAMP,
                       state = NULL
@@ -398,7 +404,8 @@ def main():
             updatedAt TIMESTAMP DEFAULT CURRENT_TIMESTAMP ON UPDATE CURRENT_TIMESTAMP NOT NULL,
             finished TIMESTAMP,
             numUpdates integer NOT NULL DEFAULT 0,
-            firstUpdate TIMESTAMP
+            firstUpdate TIMESTAMP,
+            lastUpdate TIMESTAMP
         );
     """
     create_table(conn, sql_create_projects_table)
--- a/browser/eb_application.py
+++ b/browser/eb_application.py
@@ -77,7 +77,7 @@ def action(project_id, action_type, frame):
         state.frames_changed = state.info_changed = False
 
         # Update object in local database
-        update_object(conn, (id_exists[1], state, project_id))
+        update_object(conn, state, project_id)
         conn.close()
 
     except Exception as e:
@@ -252,7 +252,7 @@ def shortcut(filename):
     return jsonify(error), 400
 
 def create_connection(_):
-    ''' Create a database connection to a SQLite database.
+    ''' Create a database connection to a MySQL database.
     '''
     conn = None
     try:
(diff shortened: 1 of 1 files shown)


browser/eb_application.py
```diff
@@ -265,6 +268,35 @@ def create_connection(_):
...
+
+        # on new server, caliban database won't exist yet
+        try:
+            query = '''CREATE DATABASE {}'''.format(config.MYSQL_DATABASE)
+            conn.cursor().execute(query)
+            conn.commit()
+
+        # it already exists
+        except:
+            pass
```
willgraf: DON'T CATCH ALL ERRORS! Just catch the expected error and pass. This could be hiding some other issues that we did NOT expect to fail.

browser/eb_application.py
```diff
@@ -150,17 +150,19 @@ def load(filename):
 
     folders = re.split('__', filename)
     filename = folders[len(folders) - 1]
-    subfolders = folders[2:len(folders)]
+    subfolders = folders[2:len(folders)-1]
```
willgraf: The linter must not be set up.

browser/eb_application.py
```diff
@@ -265,6 +268,35 @@ def create_connection(_):
         print(err)
     return conn
 
+def initial_connection(_):
```
willgraf: Why define it with args you're not using? Just delete the `_` and remove `"caliban.db"` from the places where its called.

browser/eb_application.py
```diff
@@ -265,6 +268,35 @@ def create_connection(_):
...
+        try:
+            query = '''CREATE DATABASE {}'''.format(config.MYSQL_DATABASE)
+            conn.cursor().execute(query)
+            conn.commit()
+
+        # it already exists
+        except:
+            pass
+
+        conn.close()
```
willgraf: Should put in [a `finally` block](https://docs.python.org/3/tutorial/errors.html#defining-clean-up-actions) in case there is an uncaught error in the try/except (assuming you change the error handling)

browser/eb_application.py
```diff
@@ -265,6 +268,35 @@ def create_connection(_):
...
+        conn = MySQLdb.connect(
+            user=config.MYSQL_USERNAME,
+            host=config.MYSQL_HOSTNAME,
+            port=config.MYSQL_PORT,
+            passwd=config.MYSQL_PASSWORD,
+            charset='utf8',
+            use_unicode=True)
+
+        # on new server, caliban database won't exist yet
+        try:
```
willgraf: I think two sequential try/except blocks would be better than the nested ones.

browser/eb_application.py
```diff
@@ -249,7 +252,7 @@ def shortcut(filename):
     return jsonify(error), 400
 
 def create_connection(_):
```
willgraf: Remove the useless `_` from this as well.

browser/eb_application.py
geneva-miller: Correct, no specific linter set up. I can spot fix this one (not sure which part of it needs linting in particular) and we can look into more thorough linting later?

browser/eb_application.py
geneva-miller: Set this to catch specific error (MySQLdb._exceptions.MySQLError) and print instead of pass (as with other except blocks in this file).

browser/eb_application.py
willgraf: Yea that can be a different PR but lets keep the current changes lint-appropriate too.

Issue #94, opened 2020-02-06
watershed failsafe


ngreenwald:
Watershed seems to be working great most of the time. However, the most common error mode seems to be that the two selected seed regions are not separated by a valley of lower intensity, and instead lie on a monotonically decreasing slope. Thus, when watershed is run, the point that is strictly higher ends up consuming the entire cell, leaving only a single point that is assigned to the second cell. 

This is not ideal, because it requires that the user then go in, and conversion brush that single pixel away, which must carefully be clicked on in order to be selected in the first place.

I have a couple potential solutions. The first would be to check whether or not the resulting watershed produces an object with fewer than 5 pixels, and if so to not update the labels, returning an error message instead. 

The other would be to check if that happens, and if so do a naive pixel expansion around the smaller object, so that it is at least big enough to interact with easily

geneva-miller:
#129 adds a pixel dilation if either part of the split label (old or new value) is < 5 pixels large. Between that and the ability to zoom in on the (still pretty small) part of the label, I think bad watersheds are less painful to fix now.

PR #129, opened 2020-03-29
Watershed failsafe (label dilation)

geneva-miller:
Addresses #94 across all modes (browser, desktop, trk, npz). If watershed algorithm produces only a few pixels (of either new label or old label), use dilation to expand the label to make it slightly easier to see/select. Small bugfix for desktop trk watershed included (bugfix is already incorporated into other modes but missed desktop trk class).

### DIFF
--- a/browser/caliban.py
+++ b/browser/caliban.py
@@ -19,7 +19,7 @@ import sys
 from werkzeug.utils import secure_filename
 from skimage import filters
 import skimage.morphology
-from skimage.morphology import watershed
+from skimage.morphology import watershed, dilation, disk
 from skimage.morphology import flood_fill, flood
 from skimage.draw import circle
 from skimage.measure import regionprops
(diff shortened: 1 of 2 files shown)


browser/caliban.py
```diff
@@ -972,8 +986,8 @@ def action_watershed(self, label, frame, x1_location, y1_location, x2_location,
         new_label = max(self.tracks) + 1
 
         # Locally store the frames to work on
-        img_raw = self.raw[frame]
-        img_ann = self.tracked[frame]
+        img_raw = self.raw[frame,:,:,0]
+        img_ann = self.tracked[frame,:,:,0]
```
MekWarrior: is this future-proofed against RGB? Why call out the channel specifically?

browser/caliban.py
```diff
@@ -1001,13 +1015,27 @@ def action_watershed(self, label, frame, x1_location, y1_location, x2_location,
...
+        if new_pixels < 5:
+            ws = dilation(ws, disk(3))
+
+        # ws may only leave a few pixels of old label
+        old_pixels = np.count_nonzero(ws == current_label)
+        if old_pixels < 5:
+            # create dilation image so "dimmer" label is not eroded by "brighter" label
+            dilated_ws = dilation(np.where(ws==current_label, ws, 0), disk(3))
+            ws = np.where(dilated_ws==current_label, dilated_ws, ws)
+
```
MekWarrior: Any way to easily pull this out of the class and into a Util for DRY purposes? Or should that wait for a future re-factor?

geneva-miller: I would really like to refactor the actions for DRY purposes but I think that's a future step.

browser/caliban.py
geneva-miller: Calling out channel 0 is for trk files specifically; I don't think we support multiple channels or features for trk; it was just implicit in this part of the code before. Indexing into channel 0 specifically keeps the image 2D, which helps do the dilation consistently. Otherwise, some of the dilation/np.where code would need to be reworked to use 3D arrays. Having it this way makes it consistent with npz watershed which should make a future refactor easier.

PR #139, opened 2020-04-02
Fix npz upload issues

### LINKED ISSUE #86: Browser actions sometimes don't get sent correctly
Unclear why this happens. Most of the time this is an annoying but harmless failure, and the action just needs to be tried again. However, this can also happen to file submission, so that the modified file doesn't actually get uploaded to the output s3 bucket, but in the frontend it appears to be submitted. (Also, there isn't a way to submit the file twice anyway.) We should figure out this occasional bug so that work isn't wasted.

### LINKED ISSUE #106: browser saves file locally as well as uploading to s3
I think this is a bug. Submitting a file via browser Caliban should upload it to s3 and remove it from the database, but not write the saved file to disk. I think it comes from the lines of action_save_zstack:

```
        # save secure version of data before storing on regular file system
        file = secure_filename(save_file)

        np.savez(file, raw = self.raw, annotated = self.annotated)
        path = self.subfolders
        s3.upload_file(file, self.output_bucket, path)
```

Probably, "file" should be a file-like object that is not actually written to disk.

geneva-miller:
Create health-check route and configure elastic beanstalk to use it, closes #86 (prevents file uploads from getting lost).
Upload npz and trk files to s3 from file objects, not files saved to disk, closes #106.

### DIFF
--- a/browser/caliban.py
+++ b/browser/caliban.py
@@ -16,7 +16,6 @@ import tarfile
 import tempfile
 import boto3
 import sys
-from werkzeug.utils import secure_filename
 from skimage import filters
 import skimage.morphology
 from skimage.morphology import watershed
--- a/browser/caliban.py
+++ b/browser/caliban.py
@@ -41,6 +41,7 @@ class ZStackReview:
         self.filename = filename
         self.input_bucket = input_bucket
         self.output_bucket = output_bucket
+        # subfolders is actually the full file path
         self.subfolders = subfolders
         self.rgb = rgb
         self.trial = self.load(filename)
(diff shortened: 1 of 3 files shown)


browser/eb_application.py
```diff
@@ -22,6 +22,11 @@
 app.config.from_object("config")
 
 
+@app.route("/health")
+def healthcheck():
```
willgraf, line 26: I think its usually helpful (but not required) to have the route be the same name as the function.

browser/caliban.py
```diff
@@ -601,14 +602,13 @@ def action_predict_zstack(self):
         self.create_cell_info(feature = self.feature)
 
     def action_save_zstack(self):
-        save_file = self.filename + "_save_version_{}.npz".format(self.save_version)
+        # save file to BytesIO object
+        store_npz = BytesIO()
+        np.savez(store_npz, raw = self.raw, annotated = self.annotated)
```
willgraf, line 607: Since we're changing the line, let's lint it up a bit.

Spaces go around the `=` operator UNLESS the it is inside a function call, assigning named arguments.

This line can just be `np.savez(store_npz, raw=self.raw, annotated=self.annotated)`

geneva-miller:
Figured out how to write trk files to file objects instead of to disk, so this now closes #106.

willgraf:
Approved! I suggest editing the main PR test to indicate it closes #106. GitHub does not automatically detect it from the comments.

PR #146, opened 2020-04-09
Add docker-compose.yml for easier local development

willgraf:
docker-compose is an awesome tool for running applications that rely on more than one container. This PR includes a docker-compose.yml file so that we can easily test locally with a mysql database.

- Add docker-compose.yaml
- Add basic Dockerfile for the flask app
- configure `DEBUG` as an environment variable.
- change the schema for Project from `db.String` to `db.Text`. The former gets mapped to a `VARCHAR` which requires a length, while the latter enables longer strings without a length.


browser/config.py
```diff
@@ -6,7 +6,7 @@
 from decouple import config
 
 
-DEBUG = True
+DEBUG = config('DEBUG', cast=bool, default=True)
```
geneva-miller, line 9: May want to add these to .env.example

willgraf: good call

PR #150, opened 2020-04-13
Use a <button> tag for the submit button.

### LINKED ISSUE #59: submit button should look more like a button
The button to submit a file just looks like text. This should be visually updated to more obviously be the submit button.

Also, we should consider adding a confirmation dialog to the button in case of misclick.

willgraf:
Change the HTML tag for the submit button from `<p>` to `<button>`.

A modal confirmation box would be great, but that should be a part of a migration to using a 
frontend framework.  The button does not look perfect, but it should be a good stop-gap solution until we refactor the frontend.

Screen Shot 2020-04-13 at 4 24 34 PM

---

Closes #59

### DIFF
--- a/browser/templates/index_track.html
+++ b/browser/templates/index_track.html
@@ -23,7 +23,7 @@
             <h1>Tracking Tool</h1>
             <p>Filename will be given after submit button is pressed.</p>
 
-            <button id="submit" onclick="upload()">
+            <button id="submit" type="button" onclick="upload()">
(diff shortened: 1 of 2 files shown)


geneva-miller:
A bit of the button remains after clicking it--could we make it disappear completely by changing the css in the upload function? Or something along those lines.
image

geneva-miller:
#154 makes this obsolete

Issue #142, opened 2020-04-04
Variable lag times


ngreenwald:
The amount of lag between action clicks seems to be variable in Figure8. Last night around 5 PM it was around 0.75 seconds after each action. Now it's closer 0.25 or 0.5 seconds, which is much less noticeable. 

Not sure what the difference is due to, but it makes a fairly significant difference for user experience.

willgraf:
It seems like `pngify` when `raw is True` takes up a huge amount of the processing time.

```
unpickled in 0.0011313589999986107s
pngify raw=False 0.06804040800000166s
pngify raw=True 0.2235789859999926s
got array in 0.009070264000001771s
built payload in 0.014699645999996847s
```

willgraf:
For the `handle_draw` functions, it looks like response time is dominated by the database operations, primarily the update call:

```
loaded JSON payload in 0.0003535010000064176 s
Got project from DB in 0.01933338500001014 s
Action handle_draw finished in 0.00506084299999543 s
Updated project in DB in 0.05887206099998821 s
[2020-05-07 12:21:10,087]:[DEBUG]:[application]: Action "handle_draw" for project "113" finished in 0.0869293020000157 s.
```

geneva-miller:
Re: `get_frame` I think this could be largely addressed by splitting out the cases where the raw or annotated image needs to be updated (#126). Especially for RGB mode, where the raw image never needs to be updated from the server (at least for now, while jobs are single-frame), the speed of `pngify` wouldn't change much.

Re: `handle_draw` I'll note that action time may increase as image size/amount of drawing increases, but for the image sizes that we're working with (approx 500x500 or smaller) this is not likely to be an issue, and may always be trivial compared to db operations.

willgraf:
> Re: `get_frame` I think this could be largely addressed by splitting out the cases where the raw or annotated image needs to be updated (#126). Especially for RGB mode, where the raw image never needs to be updated from the server (at least for now, while jobs are single-frame), the speed of `pngify` wouldn't change much.

Yeah the slowness was specifically for the raw RGB images, so eliminating those unnecessary calls will help. That said, `pngify` can be singificantly improved for these cases by using `PIL`:

```python
def pngify(imgarr, vmin, vmax, cmap):
    out = io.BytesIO()
    if cmap is None:
        img = Image.fromarray(imgarr)
        img.save(out, format="png")
    else:
        plt.imsave(out, imgarr,
                   vmin=vmin,
[...]

Issue #68, opened 2019-11-21
reduce lag for browser caliban


geneva-miller:
- try timing functions to identify slowest parts
- look into reducing how often object needs to be pickled/unpickled
- do connections need to be closed each time, or can we maintain a persistent connection to the db?

geneva-miller:
Addressed by #75. May be other ways we can reduce lag so I'm leaving this open for now.

willgraf:
I think using `async` functions should also help the usability here.  Won't make any of the existing functions faster, but could make the page feel a bit snappier by not locking up whenever something is executing.

ngreenwald:
Yeah, any ideas you have for this Will would be great. Currently there's about a 1/4-1/2 second lag for actions. Once you get used to it you start doing the next action before the previous one has finished, but would be nice to smooth out sometime in the future.

geneva-miller:
Maybe? But then I think there's a lot that would still need to be disabled until the updated frame is returned. I think using `async` functions could actually lead to a more confusing experience. Eg, annotator tries to watershed a label to split it, but nothing appears to happen. So they switch to pixel-editing mode and start to use the conversion brush, but partway through the labels update--what happens to the conversion brush? That type of thing is what I'm concerned about with `async`

willgraf:
> try timing functions to identify slowest parts
I've submitted #157 which logs the time it takes for most backend functions.

> do connections need to be closed each time, or can we maintain a persistent connection to the db?
Since #144 we have been allowing flask to manage the database connections, so we no longer have to worry about the

ngreenwald:
Currently there's about a 1 second lag on most actions in browser, which is slightly annoying but manageable. However, for larger images, the lag time increases significantly. This makes it prohibitively slow to work with images larger than say 250x250 pixels. 

Given that we are standardizing our image size to be 512x512, it would be great if we could speed things up enough so that images of that size can be edited in a reasonable time. This would allow us to load full images, rather than crops, once our annotators are skilled enough to use zoom.

MekWarrior:
Speed improvements are ongoing but this issue is stale. Closing in response to backend reformatting and sunsetting of js version in favor of React.

PR #165, opened 2020-05-07
Use PIL to create PNGs if not using a cmap.

willgraf:
`PIL` is much faster than `matplotlib`. I could not find a way to replace the `matplotlib` functions while still using a color map, but if no cmap is provided, we can use PIL to speed it up significantly.

Related to #68 and #142

### DIFF
--- a/browser/imgutils.py
+++ b/browser/imgutils.py
@@ -1,13 +1,20 @@
+"""Utilities for handling images"""
 import io
+
 import matplotlib.pyplot as plt
+from PIL import Image
 
 
-def pngify(imgarr, vmin, vmax, cmap):
+def pngify(imgarr, vmin, vmax, cmap=None):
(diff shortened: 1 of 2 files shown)


geneva-miller:
I'm not sure how much time this might save, but the applying a colormap could be modified to match the [desktop version.](https://github.com/vanvalenlab/caliban/blob/master/desktop/caliban.py#L1095-L1097). Then `PIL` could be used instead of `matplotlib` to save the image as a png.

willgraf:
Oh, awesome! I did not know about the `matplotlib.colors.Normalize` function. I've removed `plt.imsave` from `pngify`.

Unfortunately, it did not save much time for saving the non-raw/rgb images. It looks like <80% of the `pngify` time is in `img.save(out, format="png")` now.

PR #166, opened 2020-05-08
Don't need to return raw data on action requests.

willgraf:
This PR prevents the API from returning raw data on every action request, when actions only manipulate the annotations.  I've adapted the JS so that if the action _does_ include raw in the payload it will overwrite it currently, otherwise it just sets `rawLoaded = true`.

Are there any actions that will require an update to the raw image data?

Related to #68 and #142

### DIFF
--- a/browser/blueprints.py
+++ b/browser/blueprints.py
@@ -95,13 +95,13 @@ def action(project_id, action_type, frame):
 
     if frames_changed:
         img = state.get_frame(frame, raw=False)
-        raw = state.get_frame(frame, raw=True)
+        # raw = state.get_frame(frame, raw=True)
         edit_arr = state.get_array(frame)
 
         encode = lambda x: base64.encodebytes(x.read()).decode()
 
         img_payload = {
-            'raw': f'data:image/png;base64,{encode(raw)}',
+            # 'raw': f'data:image/png;base64,{encode(raw)}',
             'segmented': f'data:image/png;base64,{encode(img)}',
             'seg_arr': edit_arr.tolist()
         }
--- a/browser/blueprints.py
+++ b/browser/blueprints.py
@@ -80,9 +80,11 @@ def action(project_id, action_type, frame):
         # Perform edit operation on the data file
         state.action(action_type, info)
         frames_changed = state.frames_changed
+        x_changed = state._x_changed
+        y_changed = state._y_changed
         info_changed = state.info_changed
 
-        state.frames_changed = state.info_changed = False
+        state._x_changed = state._y_changed = state.info_changed = False
 
         # Update object in local database
         Project.update_project(project, state)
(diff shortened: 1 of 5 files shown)


geneva-miller:
Yes, changing the displayed channel is handled as an action, and changes only the raw image.

browser/blueprints.py
```diff
@@ -94,17 +96,17 @@ def action(project_id, action_type, frame):
     tracks = state.readable_tracks if info_changed else False
 
     if frames_changed:
-        img = state.get_frame(frame, raw=False)
-        raw = state.get_frame(frame, raw=True)
+        encode = lambda x: base64.encodebytes(x.read()).decode()
         edit_arr = state.get_array(frame)
+        img_payload = {'seg_arr': edit_arr.tolist()}
```
geneva-miller, line 101: This `seg_arr` is the array data for the annotations and does not need to be sent if `y_changed` is false.

browser/static/js/main_zstack.js
```diff
@@ -1304,7 +1304,13 @@ function action(action, info, frame = current_frame) {
         seg_array = payload.imgs.seg_arr;
 
         seg_image.src = payload.imgs.segmented;
```
geneva-miller, line 1306: Won't this image be missing if channels have been changed? Seems like this should also be loaded only `if (payload.imgs.hasOwnProperty('segmented'))`. (Same thing in main_track.js)

browser/blueprints.py
willgraf: Nice, good call. I've moved the "seg_arr" key to inside the `if y_changed` block.

browser/static/js/main_zstack.js
willgraf: Makes sense. I've added the same conditional for segmented and seg_arr for both zstack.js and track.js.

browser/static/js/main_track.js
```diff
@@ -831,9 +831,17 @@ function action(action, info, frame = current_frame) {
       if (payload.imgs) {
         // load new value of seg_array
         // array of arrays, contains annotation data for frame
-        seg_array = payload.imgs.seg_arr;
-        seg_image.src = payload.imgs.segmented;
-        raw_image.src = payload.imgs.raw;
+        if (payload.imgs.hasOwnProperty('seg_arr')) {
+          seg_arr = payload.imgs.seg_arr;
```
geneva-miller, line 835: The js variable name is `seg_array` not `seg_arr`; this should either be changed here to `seg_array` or could be changed elsewhere for consistency (same thing in main_zstack.js)

browser/static/js/main_zstack.js
```diff
@@ -1301,10 +1301,23 @@ function action(action, info, frame = current_frame) {
 
         // load new value of seg_array
         // array of arrays, contains annotation data for frame
-        seg_array = payload.imgs.seg_arr;
+        if (payload.imgs.hasOwnProperty('seg_arr')) {
+          seg_arr = payload.imgs.seg_arr;
+        }
+
+        if (payload.imgs.hasOwnProperty('segmented')) {
+          segLoaded = false;
```
geneva-miller, line 1309: I think we can just set `segLoaded` to `false` in this block, and remove line 1300 as well as the `else` case for loading images. (Similar thing to loading raw--we only need to set `rawLoaded` to `false` if we're changing the image source, otherwise it remains `true`)

browser/static/js/main_track.js
willgraf: good catch!

browser/caliban.py
```diff
@@ -743,9 +748,14 @@ def __init__(self, filename, input_bucket, output_bucket, subfolders):
 
         self.current_frame = 0
 
-        self.frames_changed = False
+        self._x_changed = False
+        self._y_changed = False
         self.info_changed = False
 
+    @property
```
geneva-miller, line 755: no longer needed

PR #168, opened 2020-05-12
Save npzs with X and y

### LINKED ISSUE #160: return npz with 'X' and 'y' instead of 'raw' and 'annotated'
Currently caliban has the old nomenclature for the created NPZs. To standardize the pipeline it would be nice to switch to X/y, or at least use whichever name the input npz has

geneva-miller:
Close #160. For historical reasons we may have npzs still floating around saved with variable names `raw` and `annotated` instead of `X` and `y`, so these npzs should still be loaded by caliban, but we have no particular reason not to save them as `X` and `y` going forward.

### DIFF
--- a/browser/caliban.py
+++ b/browser/caliban.py
@@ -627,6 +627,8 @@ class ZStackReview:
     def action_save_zstack(self):
         # save file to BytesIO object
         store_npz = io.BytesIO()
+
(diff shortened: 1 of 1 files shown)


geneva-miller:
That would be interesting to explore; I've opened #169 as a reminder to look into this.

PR #174, opened 2020-06-08
Hide labels, bugfixes

### LINKED ISSUE #173: optionally hide labels in pixel-editing mode
Useful viewing option if annotators are restricted to non-rgb pixel-only mode. Specifically helps with looking at edges of cells. Probably just "L" keybind for now in non-rgb mode (I don't think there's a need for this for rgb browser mode?) and could add a checkbox in the future if we figure out an appropriate place to put it. Would be nice to eventually have this be an annotation opacity slider instead of a binary toggle.

geneva-miller:
Adds option, in non-rgb pixel-only mode, to hide labels with "L" key, as users can't leave pixel-editing mode to view the raw-only image. Doesn't change existing rgb "L" functionality. Adds intermediate step to image processing cascade so that the adjusted, un-composited raw image can be easily displayed when "L" is toggled. Closes #173

Also addresses small bugs encountered while making instructions for HeLa movies:
- thresholding failed if no labels existed, due to edge case behavior when determining the maximum label value. Didn't seem to affect anything other than thresholding. Fixed by specifically accounting for that edge case
- setting the conversion brush with clicking would register the click locations as part of the `mouse_trace` and send those click coordinates next time the conversion brush was used. Fixed by not pushing coordinates to `mouse_trace` if in `prompt` mode (as opposed to `drawing` mode).

### DIFF
--- a/browser/static/js/adjust.js
+++ b/browser/static/js/adjust.js
@@ -138,8 +138,8 @@ function preCompAdjust() {
   preCompSeg.src = canvas.toDataURL();
 }
 
-// adjust raw further (if needed), composite annotations on top
-function compositeImages() {
+// adjust raw further, pre-compositing (use to draw when labels hidden)
+function preCompRawAdjust() {
   let canvas = document.getElementById('hidden_seg_canvas');
   let ctx = $('#hidden_seg_canvas').get(0).getContext('2d');
   ctx.imageSmoothingEnabled = false;
@@ -153,6 +153,17 @@ function compositeImages() {
     invert(rawData);
   }
   ctx.putImageData(rawData, 0, 0);
+
+  preCompRaw.src = canvas.toDataURL();
+}
+
+// composite annotations on top of adjusted raw image
+function compositeImages() {
+  let canvas = document.getElementById('hidden_seg_canvas');
+  let ctx = $('#hidden_seg_canvas').get(0).getContext('2d');
+  ctx.imageSmoothingEnabled = false;
+
+  ctx.drawImage(preCompRaw, 0, 0, rawWidth, rawHeight);
 
   // add labels on top
   ctx.save();
--- a/browser/static/js/main_zstack.js
+++ b/browser/static/js/main_zstack.js
@@ -87,7 +87,7 @@ class Mode {
       brightness = 0;
       current_contrast = 0;
       prepareRaw();
-    } else if ((key === 'l' || key === 'L') && !edit_mode) {
+    } else if ((key === 'l' || key === 'L') && rgb && !edit_mode) {
       display_labels = !display_labels;
       render_image_display();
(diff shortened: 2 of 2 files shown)


browser/static/js/main_zstack.js
```diff
@@ -1144,8 +1156,8 @@ function handle_scroll(evt) {
     let mod_contrast = -Math.sign(evt.originalEvent.deltaY) * 4;
     // stop if fully desaturated
     current_contrast = Math.max(current_contrast + mod_contrast, -100);
-    // stop at 5x contrast
-    current_contrast = Math.min(current_contrast + mod_contrast, 400);
+    // stop at 8x contrast
+    current_contrast = Math.min(current_contrast + mod_contrast, 700);
```
willgraf, line 1160: Instead, we can define global variables `const MAX_CONTRAST = 700` and `MAX_CONTRAST = -100` to make this easier to configure.

browser/static/js/adjust.js
```diff
@@ -138,8 +138,8 @@ function preCompAdjust() {
   preCompSeg.src = canvas.toDataURL();
 }
 
-// adjust raw further (if needed), composite annotations on top
-function compositeImages() {
+// adjust raw further, pre-compositing (use to draw when labels hidden)
+function preCompRawAdjust() {
```
willgraf, line 142: If possible, we should edit this function so that it doesn't assume any global variables.  If this function is called in the wrong order, it may `rawData`, `rawWidth`, `rawHeight`, or `preCompRaw` may have unexpected values.

Instead, I think we can edit this function (and eventually all of the functions) so that they use as few out-of-scope variables as possible.

geneva-miller: I'd like to address this in a separate PR to refactor `adjust.js`, if that seems appropriate to you.

willgraf: I don't think its best to commit a function that already has a TODO in it. That said, maybe it is appropriate to wait for the `adjust.js` refactor; I don't know what that entails.

The more functions we add that use these out-of-scope variables, the harder it is to effectively write new code, and the more convoluted the frontend logic becomes.

geneva-miller: Right, many of the image processing functions in `adjust.js` (the ones that load an image, apply some processing steps, then save the adjusted image to the `src` of an `Image` for later use) also use out-of-scope variables. The change in this PR does split one of those functions into two and adds a few more references to `rawData`, `rawWidth` etc that could be fixed in this PR. I'm suggesting that we could address all the `adjust.js` functions with out-of-scope variables in a separate PR intended for that, instead of fixing `preCompRawAdjust` here and getting the rest later.

I'm open to how you would like me to fix the out-of-scope variables, I just wanted to clarify what I meant by an `adjust.js` refactor.

PR #176, opened 2020-06-12
Browser/refactor/adjust variable scope

geneva-miller:
A lot of functionality in browser caliban currently relies on access to global variables, which is not ideal. This PR will rescope the image adjustment functions to reduce the amount of code that depends on global variables. We may identify other ways to refactor the image adjustment functions to make them more modular.


geneva-miller:
I think we could pull these functions into a module or class pattern. The image data manipulation functions could remain as private methods while other functions are exposed to the rest of the caliban code. I think we'd be able to then pass in `rawWidth` and `rawHeight` upon initialization, since these correspond to the true image size and will not change once the file is opened. I think we could also make the `hidden_seg_canvas` belong to `adjust.js` in the same way that `brush.js` uses its own canvas for the brush-specific image manipulation steps. @willgraf do you have any thoughts or preferences on how to structure that?

geneva-miller:
Thinking about other things that would make sense in the `ImageAdjustment` class. Contrast and brightness values (and methods)? The Image objects that store the images for later display (could see about moving Image `onload` behavior, loading/processing cascade all into the `ImageAdjustment` class)?

browser/static/js/main_zstack.js
```diff
@@ -1395,26 +1410,35 @@ function start_caliban(filename) {
     handle_mouseup();
    });
 
-  // define image onload cascade behavior
+  load_file(filename);
+
+  // define image onload cascade behavior, need rawHeight and rawWidth first
+  adjust = new ImageAdjustment(width=rawWidth, height=rawHeight)
```
willgraf, line 1416: Refactor the object's variable name to be a noun, `adjuster` maybe. I would refactor the class to something like `ImageAdjster` too. IMO `ImageAdjustment` sounds like a class that doesn't get instantiated.

browser/static/js/adjust.js
```diff
@@ -1,277 +1,290 @@
...
-function segAdjust() {
-  segLoaded = true;
-  if (rawLoaded && segLoaded) {
-    if (rgb) {
-      postCompAdjustRGB();
-    } else {
-      compositeImages();
+  // TODO: not sure how to scope segLoaded and rawLoaded.
+  // they are changed by these adjust functions so the image composite
+  // will only be generated when both source images are ready to use
```
willgraf, line 264: It looks like `segLoaded` and `rawLoaded` are only used in these 2 methods. Can these functions get pulled out of the class and moved into the place where it's being called? That would remove the troublesome variables from `adjust.js` completely.

browser/static/js/main_zstack.js
```diff
@@ -732,9 +747,11 @@ let mousedown = false;
 let spacedown = false;
 var tooltype = 'draw';
 var project_id;
-var brush;
 let mouse_trace = [];
 
+var brush;
+var adjust;
```
willgraf, line 753: If these are only ever assigned once, we can set them to `const` instead of `var`.

browser/static/js/adjust.js
```diff
@@ -1,277 +1,290 @@
...
+          this._recolorScaled(ann, i, j, jlen, r=255, g=-255, b=-255);
+          continue;
+        // outline white single
+        } else if (singleOutline && currentVal === -o1) {
+          this._recolorScaled(ann, i, j, jlen, r=255, g=255, b=255);
+          continue;
+        // outline all remaining edges with white
+        } else if (outlineAll && currentVal < 0) {
+          this._recolorScaled(ann, i, j, jlen, r=255, g=255, b=255);
+          continue;
```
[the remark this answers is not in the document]
geneva-miller: Both of those cases recolor edges to white, but are used in different situations. `_recolorScaled` recolors a single pixel. So when `outlineAll` is true, everything gets an outline (label edges are indicated in the label array by a negative label)--that's the part used for RGB image overlays. But in some cases (conversion brush in non-rgb mode) only the edges of one particular label should be visualized with white. They are pretty similar though.

browser/static/js/adjust.js
geneva-miller: Yes--I think it'll work to make `segLoaded` an attribute of the `ImageAdjuster` class and set it to false at the start of `preCompAdjust`. (Similarly with `rawLoaded`.) Then the only place in the main js file that it needs to be explicitly updated is when new images are loaded (changed frames, modified annotations, changed channels).

browser/static/js/adjust.js
```diff
@@ -1,277 +1,290 @@
...
+    // draw rawImage so we can extract image data
+    this.ctx.clearRect(0, 0, this.width, this.height);
+    this.ctx.drawImage(rawImage, 0, 0, this.width, this.height);
+    let rawData = this.ctx.getImageData(0, 0, this.width, this.height);
+    this._contrast_image(rawData, contrast, brightness);
+    this.ctx.putImageData(rawData, 0, 0);
 
-  contrastedRaw.src = canvas.toDataURL();
-}
+    contrastedRaw.src = this.canvas.toDataURL();
```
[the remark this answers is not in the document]
geneva-miller: Okay, so more along the lines of
```
preCompAdjust () {
  // function body
return canvas.toDataURL();
}
```
?
And then within the main js code, instead of calling `adjuster.preCompAdjust(preCompSeg);` it's more like
```
preCompSeg.src = adjuster.preCompAdjust();
```

browser/static/js/adjust.js
```diff
@@ -1,277 +1,290 @@
...
-      d[i + 2] = d[i+2]*contrast + brightness;
+  // modify image data in place to recolor
+  _recolorScaled(data, i, j, jlen, r=255, g=255, b=255) {
+    // location in 1D array based on i and j
+    let pixel_num = jlen*j + i;
+    // set to color by changing RGB values
+    // data is clamped 8bit type, so +255 sets to 255, -255 sets to 0
+    data[(pixel_num*4)] += r;
+    data[(pixel_num*4) + 1] += g;
+    data[(pixel_num*4) + 2] += b;
```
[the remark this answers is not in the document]
geneva-miller: Huh, yeah, it looks like `_contrast_image`, `_grayscale`, and `_invert` each return the image, but I don't think that returned object even gets used by the functions that call them, they just work in place. I think it should be pretty simple to change them to only working in place for consistency.

browser/static/js/main_zstack.js
geneva-miller: Currently this is still tricky to do. The `ImageAdjuster` class needs to be initialized with height and width attributes, which we don't know until information about the specific file is loaded. So that gets initialized in the `start_caliban` function. But the `adjuster` object still needs to have global scope (for now) to play nice with everything else, so that variable gets declared outside of `start_caliban`, which won't work with `const`. So I'm not sure we can easily make that switch until other big pieces of the code get refactored as well.

browser/static/js/adjust.js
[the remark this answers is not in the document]
geneva-miller: Actually, while I think that approach would work for most of the places these functions are called, it won't work as well within `start_caliban`:
`    rawImage.onload = () => adjuster.contrastRaw(rawImage, contrastedRaw, current_contrast, brightness);`
Not sure how to address this to nicely handle both use cases.

geneva-miller:
The `mode` and `brush` objects that are passed in as args to some of the functions are because a handful of attributes need to be accessed. For example, `postCompAdjust` checks `brush.conv`, `brush.target`, and `brush.value`. I'm not sure about internalizing `mode` and `brush` to `ImageAdjuster`, at least in this round of refactoring. I think there's a few more areas where it would make sense to refactor into classes/reduce global variable scope before we solidify how different classes interact with each other.

browser/static/js/adjust.js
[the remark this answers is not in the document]
willgraf: The `if` conditions are different, but both do the same `_recolorScaled` call.

Not a big deal if there are different conditions that want the same result, but i just want to confirm that both conditions ARE supposed to do the same thing: `this._recolorScaled(ann, i, j, jlen, r=255, g=255, b=255);`

browser/static/js/adjust.js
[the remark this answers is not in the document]
willgraf: Would that code not work?

I haven't run anything, but the code you have right there seems like it would work.

browser/static/js/adjust.js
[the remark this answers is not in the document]
geneva-miller: Ah, yes, both of those if cases are supposed to result in that call, which recolors the selected pixel to white.

browser/static/js/adjust.js
```diff
@@ -1,277 +1,376 @@
...
+      this.contrastedRaw.onload = () => this.rawAdjust(current_highlight, edit_mode, brush, mode);
+      this.segImage.onload = () => this.preCompAdjust(current_highlight, edit_mode, brush, mode);
+      this.preCompSeg.onload = () => this.segAdjust(current_highlight, edit_mode, brush, mode);
+    } else {
+      this.rawImage.onload = () => this.contrastRaw();
+      this.contrastedRaw.onload = () => this.preCompRawAdjust();
+      this.preCompRaw.onload = () => this.rawAdjust(current_highlight, edit_mode, brush, mode);
+      this.segImage.onload = () => this.preCompAdjust(current_highlight, edit_mode, brush, mode);
+      this.preCompSeg.onload = () => this.segAdjust(current_highlight, edit_mode, brush, mode);
+      this.compositedImg.onload = () => this.postCompAdjust(edit_mode, brush);
```
willgraf, line 72: `current_highlight`, `edit_mode`, `brush`, and `mode` are all undefined.

browser/static/js/adjust.js
```diff
@@ -1,277 +1,376 @@
...
-  for (var j = 0; j < seg_array.length; j += 1){ //y
-    for (var i = 0; i < seg_array[j].length; i += 1){ //x
-      let jlen = seg_array[j].length;
-      let currentVal = Math.abs(seg_array[j][i]);
-      if (currentVal === h1 || currentVal === h2) {
-        recolorScaled(ann, i, j, jlen, r=255, g=-255, b=-255);
+    // brightness and contrast adjustment
+    // TODO: have this returned by a getter w/o a setter?
+    this.MIN_CONTRAST = -100;
+    this.MAX_CONTRAST = 700;
```
willgraf, line 33: I think the getter is a great idea. If a getter is used, this variable should be refactored to `_maxContrast` or something.

browser/static/js/adjust.js
```diff
@@ -1,277 +1,376 @@
...
-  preCompSeg.src = canvas.toDataURL();
-}
+  preCompositeLabelMod(img, h1, h2) {
+    let r, g, b;
+    let ann = img.data;
+    // use label array to figure out which pixels to recolor
+    for (var j = 0; j < seg_array.length; j += 1){ // y
+      for (var i = 0; i < seg_array[j].length; i += 1){ // x
+        let jlen = seg_array[j].length;
+        let currentVal = Math.abs(seg_array[j][i]);
```
willgraf: `seg_array` is undefined.

browser/static/js/adjust.js
```diff
@@ -1,277 +1,376 @@
...
+      outlineAll = false,
+      translucent=false, t1=-1, t2=-1) {
+
+    let r, g, b;
+    let ann = img.data;
+    // use label array to figure out which pixels to recolor
+    for (var j = 0; j < seg_array.length; j += 1){ // y
+      for (var i = 0; i < seg_array[j].length; i += 1){ // x
+        let jlen = seg_array[j].length;
+        let currentVal = seg_array[j][i];
```
willgraf: `seg_array` is undefined.

browser/static/js/adjust.js
```diff
@@ -1,277 +1,376 @@
...
+  compositeImages() {
 
-  ctx.putImageData(imgData, 0, 0);
+    this.ctx.drawImage(this.preCompRaw, 0, 0, this.width, this.height);
 
-  postCompImg.src = canvas.toDataURL();
-}
+    // add labels on top
+    this.ctx.save();
+    this.ctx.globalAlpha = 0.3;
```
willgraf, line 276: What is this number? Should it be paramterized?

browser/static/js/adjust.js
geneva-miller: Those variables are still globally scoped--I think we will need to iterate on how these pieces interact with each other as the code becomes more modular.

browser/static/js/adjust.js
geneva-miller: That determines how transparent labels should be when they're drawn on top of the raw image to make the pixel-editing overlay image. Parameterizing it sounds good, especially since it would be nice (eventually) to have it be an adjustable value.

browser/static/js/adjust.js
willgraf: Since they are all only used in the constructor, they could just be passed in.

geneva-miller: I gave this a try, but passing in the variables in the constructor doesn't result in the correct behavior.

geneva-miller: I think in the future, we may be able to store relevant settings in `adjuster` attributes, and have them update when the settings change. But I think that's out of scope for this PR and will likely depend on other parts of the code (especially `mode`) being refactored first so it doesn't get entirely out of hand.

browser/static/js/adjust.js
```diff
@@ -1,277 +1,398 @@
...
+  // TODO: do we want to pass in B&C values or use object attributes?
+  _contrastImage(img, contrast = 0, brightness = 0) {
+    const d = img.data;
+    contrast = (contrast / 100) + 1;
+    for (let i = 0; i < d.length; i += 4) {
+      d[i] = d[i] * contrast + brightness;
+      d[i + 1] = d[i + 1] * contrast + brightness;
+      d[i + 2] = d[i + 2] * contrast + brightness;
+    }
+  }
```
willgraf, line 169: I think we can leave it as taking inputs instead of attributes
```suggestion
  _contrastImage(img, contrast = 0, brightness = 0) {
    const d = img.data;
    contrast = (contrast / 100) + 1;
    for (let i = 0; i < d.length; i += 4) {
      d[i] = d[i] * contrast + brightness;
      d[i + 1] = d[i + 1] * contrast + brightness;
      d[i + 2] = d[i + 2] * contrast + brightness;
    }
  }
```

browser/static/js/adjust.js
```diff
@@ -1,277 +1,398 @@
...
+  get maxContrast() {
+    return this._maxContrast;
+  }
+
+  // TODO: getters/setters for brightness and contrast?
+  // or use this group of methods?
+
+  changeContrast(inputChange) {
+    const modContrast = -Math.sign(inputChange) * 4;
+    // stop if fully desaturated
```
willgraf, line 108: I think this TODO has been resolved by the getters above.

```suggestion
  changeContrast(inputChange) {
    const modContrast = -Math.sign(inputChange) * 4;
    // stop if fully desaturated
```

PR #181, opened 2020-07-12
Desktop/input data shapes

geneva-miller:
Allows Caliban to open 3D npzs as single frame npzs (or 5D npzs as long as first dimension can be squeezed out). Refactors Mode into specific subclasses to provide more accurate question/prompt text displays. Ie, files that have only one frame don't need to have references to frame number in display text.


desktop/mode.py
```diff
@@ -50,115 +51,208 @@ def update(self, kind, **info):
...
+
+        elif self.action == "DRAW BOX":
+            self.text = "\nDraw a bounding box around the area you want to threshold."
+
+        elif self.action == "CONVERSION BRUSH TARGET":
+            self.text = "\nClick on the label you want to draw OVER."
+
+        elif self.action == "CONVERSION BRUSH VALUE":
+            self.text = ("\nClick on the label you want to draw WITH,"
+            " or press N to set the brush to an unused label.")
```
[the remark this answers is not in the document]
willgraf: The PEP8 spacing could be fixed in a lot of places.

desktop/mode.py
```diff
@@ -50,115 +51,208 @@ def update(self, kind, **info):
...
-                    "\nS = PREDICT THIS FRAME FROM PREVIOUS FRAME"
-                    "\nSPACE = PREDICT ALL FRAMES"
-                    "\nESC = CANCEL")
+            elif self.action == "SWAP":
+                self.text = ("\nSwap {} & {}?\n"
+                        + self.simple_answer).format(self.label_1, self.label_2)
+
+            elif self.action == "CREATE NEW":
+                self.text = ("\nCreate new label from label {0}?\n"
+                        + self.simple_answer).format(self.label)
```
willgraf, line 195: why not just format `self.simple_answer` too instead of concatenating it.

desktop/mode.py
```diff
@@ -50,115 +51,208 @@ def update(self, kind, **info):
...
-                    "\nESC = CANCEL")
+                self.text = ("\nRelabel annotations?\n"
+                        + self.simple_answer)
 
-        elif self.kind == "PROMPT":
-            if self.action == "FILL HOLE":
-                text = "\nSelect hole to fill in label {}.".format(self.label)
+    @staticmethod
+    def none():
+        return Mode2D(None)
```
willgraf, line 203: Is there anyway to bake this into the parent class so we don't have to redefine it in each child class?

desktop/mode.py
```diff
@@ -29,7 +29,8 @@ def __init__(self, kind, **info):
         self.kind = kind
```
willgraf, line 29: I am not a fan of this variable name bc I cannot tell what it is for.

desktop/caliban.py
```diff
@@ -4684,14 +4687,25 @@ def save(self):
...
+        # make sure has same dims as original
+        if self.dims == 3:
+            raw = np.squeeze(self.raw, axis=0)
+            ann = np.squeeze(self.raw, axis=0)
+        elif self.dims == 4:
+            raw = self.raw
+            ann = self.annotated
+        elif self.dims == 5:
+            raw = np.expand_dims(self.raw, axis=0)
+            ann = np.expand_dims(self.annotated, axis=0)
```
willgraf, line 4699: I've seen this little pattern before. Would it make sense to have a helper method that casts to the right dimension?

desktop/mode.py
geneva-miller: it's a legacy variable name with possible values "SELECTED", "MULTIPLE", "QUESTION", "PROMPT", and "DRAW". I'm open to alternative names although I'll note that this variable is everywhere in both desktop and browser

willgraf: Maybe it's not worth changing in this PR but it definitely seems like it could be renamed for clarity. We can chew on it and make another PR in the future.

geneva-miller: I think that could be a good opportunity to also turn the fixed value strings into class attributes.

desktop/caliban.py
geneva-miller: I'm not sure. I don't anticipate using this pattern very broadly so I think it would be fine to keep as is for now

geneva-miller:
Opened #183 for the `Mode` structural changes that won't be implemented in this PR

desktop/caliban.py
```diff
@@ -3109,10 +3104,17 @@ def __init__(self, filename, raw, annotated, save_vars_mode):
         try:
             first_key = list(self.cell_info[0])[0]
             display_info_types = self.cell_info[0][first_key]
-            self.display_info = [*sorted(set(display_info_types) - {'frames'})]
+            if self.single_frame:
+                self.display_info = list((set(display_info_types) - {'frames'} - {'slices'}))
```
willgraf, line 3108: I think this can be a single set.
```suggestion
                self.display_info = list((set(display_info_types) - {'frames', 'slices'}))
```

desktop/mode.py
```diff
@@ -50,115 +51,208 @@ def update(self, kind, **info):
...
-                        "\nESC = CANCEL").format(self.label_2, self.label_1)
+        elif self.kind == "PROMPT":
+            self.set_prompt_text()
 
-            elif self.action == "PARENT":
-                text = "\nMake {} a daughter of {}?\n{}".format(self.label_2, self.label_1, answer)
+        elif self.kind == "DRAW":
+            self.text = ("\nUsing conversion brush to replace {} with {}."
+                         "\nUse ESC to stop using the conversion brush.").format(
+                self.conversion_brush_target, self.conversion_brush_value)
```
willgraf, line 164:
```suggestion
        elif self.kind == "DRAW":
            self.text = ("\nUsing conversion brush to replace {} with {}."
                         "\nUse ESC to stop using the conversion brush.").format(
                             self.conversion_brush_target, self.conversion_brush_value)
```

geneva-miller: I'm not sure about this one. The multiple strings definitely fall under "need to be aligned with opening delimiter" but I think the `.format` hanging indent applies to the indentation level of `self.text`, not the string. I think we can leave it as is and make that modification later, once we start linting with pytest, if we need.

willgraf: OK no worries

desktop/caliban.py
```diff
@@ -3887,18 +3886,19 @@ def label_mode_question_keypress_helper(self, symbol, modifiers):
...
+            if not self.single_frame:
+                if symbol == key.U:
+                    self.action_relabel_unique()
+                    self.mode.clear()
+                if symbol == key.P:
+                    self.action_relabel_preserve()
+                    self.mode.clear()
+                if symbol == key.S:
+                    self.action_relabel_frame()
+                    self.mode.clear()
```
MekWarrior, line 3901: These nested if blocks are very concerning. At some point, we're going to have to pay this technical debt. We are adding yet another conditional with this PR, and we need to call that out.

geneva-miller: Yeah, I don't love how keypresses are handled. This is definitely something I've been turning over in the back of my mind. I think this will be tied into or depend on a deeper refactor of `Mode` and may also involve creating specific event handlers https://pyglet.readthedocs.io/en/latest/programming_guide/events.html

geneva-miller: For this PR, though, I want to make sure there aren't hidden keypress options. Many of the single-frame actions wouldn't change the outcome of the action, but if the prompt indicates space to confirm/esc to cancel, I want space to be the only way to confirm. Functionally, I think the most important one of those is saving, as saving a single frame npz as a trk is totally useless.

desktop/caliban.py
```diff
@@ -3092,8 +3076,19 @@ def __init__(self, filename, raw, annotated, save_vars_mode):
...
+        # should be robust to 3D, 4D, and some cases of 5D array sizes
+        self.dims = raw.ndim
+        if self.dims == 3:
+            self.raw = np.expand_dims(self.raw, axis=0)
+            self.annotated = np.expand_dims(self.annotated, axis=0)
+
+        elif self.dims == 5:
+            self.raw = np.squeeze(self.raw, axis=0)
+            self.annotated = np.squeeze(self.annotated, axis=0)
+
```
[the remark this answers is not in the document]
willgraf: Yeah this makes sense, we should probably let the user know if we're making assumptions about their data.

desktop/caliban.py
willgraf: Can you please add this TODO to the Mode refactor issue, which should get bumped way up the priority list.

desktop/caliban.py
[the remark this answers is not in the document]
geneva-miller: "Raw and annotated images must both be in 4D arrays in the shape (frames, y, x, channels or features)." is right at the top of the README--so I do feel like we are letting them know we've made that assumption of the data. 3D/5D opening is more to catch situations where a user got this wrong due to a trivial dimension (I've done this before when breaking up larger datasets into npzs for Caliban). I don't think we need to add additional messaging to the program itself, although I wouldn't be opposed to slightly modifying that part of the README.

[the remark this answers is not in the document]
MekWarrior: Not all users read docs or REAMEs, and it would still be a useful alert. In your case, it would have let you know you've broken up the dataset incorrectly, which could cause other errors downstream. No need to change the README, but we should alert the user when we transform data. A prompt may be too much, but a logging statement would be appropriate.

PR #190, opened 2020-07-23
Create BaseReview class to better share similar functionality.

willgraf:
Add a parent class for `TrackReview` and `ZStackReview` so they can better share code and functionality. While `cell_info` and `tracks` can likely be better aligned, I think this represents a good checkpoint as far as DRYing out the browser code.

- [ ] tests for `caliban.py`
- [ ] make `cell_info` and `tracks` the same field
- [ ] change `ZStackReview` to rely on only `cell_info` and not `cell_ids` to better align the classes.


browser/caliban.py
```diff
@@ -22,93 +23,70 @@
...
+        self.channel_max = self.raw.shape[-1]
+        self.feature_max = self.annotated.shape[-1]
+        self.max_frames = self.raw.shape[0]
+        self.height = self.raw.shape[1]
+        self.width = self.raw.shape[2]
 
-        for feature in range(self.feature_max):
-            self.create_cell_info(feature)
+        self._x_changed = False
+        self._y_changed = False
```
geneva-miller, line 62: `self._y_changed` is initialized in line 42, `self._x_changed` should be moved there or `self.info_changed` should be moved here.

browser/caliban.py
```diff
@@ -363,237 +400,195 @@ def action_threshold(self, y1, x1, y2, x2, frame, label):
...
+            raw_key='raw', annotated_key='annotated')
 
-        in_modified = np.any(np.isin(filled_img_ann, old_label))
+        self.rgb = rgb
 
-        # update cell info dicts since labels are changing
-        self.add_cell_info(feature=self.feature, add_label=new_label, frame=frame)
+        if self.rgb:
+            # possible differences between single channel and rgb displays
+            if self.raw.ndim == 3:
```
geneva-miller, line 418: this happens after `self.width` and `self.height` are assigned and could cause problems. Maybe reassign the size attributes if the dimensions are updated? Otherwise I would expect a (512, 512, 3) image to be have "width" of 2 and 512 "frames".

browser/caliban.py
```diff
@@ -699,63 +684,30 @@ def create_cell_info(self, feature):
...
 
             for frame in range(self.annotated.shape[0]):
-                if cell in annotated[frame, :, :]:
+                if cell in annotated[frame, ...]:
                     self.cell_info[feature][cell]['frames'].append(int(frame))
             self.cell_info[feature][cell]['slices'] = ''
 
         self.info_changed = True
 
-    def create_lineage(self):
```
geneva-miller, line 708: `create_lineage` is used to save npzs as trk files. currently not in use for browser version so it's fine to take out in this PR, but I think we'll add it back in to make a shared backend for browser/desktop at some point

willgraf: Yeah, we can definitely add it back in the future but it did not run in its current state.

browser/caliban.py
willgraf: Would just assigning the width and height again resolve the problem?

```python
    if self.raw.ndim == 3:
            self.raw = np.expand_dims(self.raw, axis=0)
            self.annotated = np.expand_dims(self.annotated, axis=0)

            self.rgb_img = self.reduce_to_RGB()

            self.max_frames = self.raw.shape[0]
            self.height = self.raw.shape[1]
            self.width = self.raw.shape[2]
```

geneva-miller: Yes, I think that would fix the issue.

PR #191, opened 2020-07-25
Cleaned up the JS a bit using ESLint

willgraf:
Reading through I noticed a few minor code cleanup changes to make, and used a local ESLinter to clean up everything I could. No functionality was changed, but below is a summary of edits.

- snake_case variables were transformed to camelCase
- fixed any whitespace issues (with `=` operators and added space between `function {`.
- changed any once-assigned variables to `const`.
- Use single quotes `'` instead of double quotes `"`.
- Added a couple more TODOs, mostly about assumptions of existing variables.

### DIFF
--- a/browser/static/js/adjust.js
+++ b/browser/static/js/adjust.js
@@ -1,6 +1,6 @@
 // helper functions
 
-class ImageAdjuster{
+class ImageAdjuster {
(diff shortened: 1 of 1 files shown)


geneva-miller:
Thank you, that's a lot nicer!

PR #194, opened 2020-07-30
Prevent desktop caliban crash if file can't be saved

### LINKED ISSUE #177: desktop caliban crashes if saving isn't successful
Mostly happens if trying to write the saved file in a folder without correct permissions. This is a problem because then the work you were trying to save is immediately lost. Exception handling would probably be a good start here.

geneva-miller:
Closes #177 by adding try/except statements to file saving functions.


geneva-miller:
Thanks, I'll make a note of the trk saving and keep it in mind for future changes.

PR #195, opened 2020-07-30
Browser/bugfix/update trk js

geneva-miller:
Fixes trk javascript functionality (brush displays and draws correctly now). Doesn't update trk js to use the modules that the zstack js uses, just patches things for now.

### DIFF
--- a/browser/static/js/main_track.js
+++ b/browser/static/js/main_track.js
@@ -640,7 +640,7 @@ function render_edit_image(ctx) {
   ctx.restore();
 
   ctx.save();
-  let region = new Path2D();
+  const region = new Path2D();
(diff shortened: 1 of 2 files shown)


browser/static/js/main_track.js
```diff
@@ -638,8 +639,15 @@ function render_edit_image(ctx) {
   ctx.drawImage(seg_image, padding, padding, dimensions[0], dimensions[1]);
   ctx.restore();
 
+  ctx.save();
+  let region = new Path2D();
```
willgraf, line 643: since you're not reassigning `region`, you can use `const`.

```suggestion
  const region = new Path2D();
```

PR #196, opened 2020-08-09
Cursor, viewing window js classes

### LINKED ISSUE #158: handle_draw sends duplicate mouse trace coordinates in payload.

We should both:

geneva-miller:
Make browser js code more modular (and reduce globally scoped variables) by grouping some functions and variables into classes. This PR introduces a cursor class to keep track of mouse position on canvas (and related information and methods), and a viewing window class to handle panning, zooming, drawing borders around, and keeping track of the currently displayed portion of the image.

Originally intended to substantially clean up mouse events, but these may be trickier as they are entangled with `Mode` behavior. Hoping to close #158 by refining the behavior of `mouse_trace` (now `cursor.trace`).

### DIFF
--- a/browser/static/js/cursor.js
+++ b/browser/static/js/cursor.js
@@ -11,6 +11,9 @@ class CalibanCursor {
     // mouse position on canvas, no adjustment for padding
     this.rawX = 0;
     this.rawY = 0;
+    // adjusted for padding
+    this.canvasPosX = -5;
+    this.canvasPosY = -5;
   }
 
   // check if the mouse position in canvas matches to a displayed part of image
--- a/browser/static/js/cursor.js
+++ b/browser/static/js/cursor.js
@@ -6,6 +6,11 @@ class CalibanCursor {
     this.scale = scale;
     this.pressed = false;
     this.trace = [];
+
+    // mouse coords
+    // mouse position on canvas, no adjustment for padding
+    this.rawX = 0;
+    this.rawY = 0;
   }
 
   // check if the mouse position in canvas matches to a displayed part of image
--- a/browser/static/js/cursor.js
+++ b/browser/static/js/cursor.js
@@ -7,6 +7,17 @@ class CalibanCursor {
     this.pressed = false;
     this.trace = [];
   }
+
+  // check if the mouse position in canvas matches to a displayed part of image
+  inRange(x, y) {
+    if (x >= 0 && x < this.width * this.scale &&
+        y >= 0 && y < this.height * this.scale) {
+      return true;
+    } else {
+      return false;
+    }
+  }
+
 }
 
 // handle updating zooming and panning attributes
--- a/browser/static/js/cursor.js
+++ b/browser/static/js/cursor.js
@@ -1,3 +1,7 @@
+// NOTE: cursor and viewer are more abstract than the brush,
+// how they should relate to each other is an open question.
+// eg, should scale/dimensions be attributes only of viewer? etc
+
 // cursor to handle mouse movement and attributes
 class CalibanCursor {
   constructor(width, height, scale) {
--- a/browser/static/js/cursor.js
+++ b/browser/static/js/cursor.js
@@ -2,6 +2,7 @@
 class CalibanCursor {
   constructor() {
     this.pressed = false;
+    this.trace = [];
   }
 }
 
--- a/browser/static/js/cursor.js
+++ b/browser/static/js/cursor.js
@@ -35,7 +35,84 @@ class CanvasView {
 
     this.scale = 1;
 
+    this.topBorder = new Path2D();
+    this.bottomBorder = new Path2D();
+    this.rightBorder = new Path2D();
+    this.leftBorder = new Path2D();
+
   }
+
+  setBorders(padding) {
+    const scaledWidth = this.scale * this.width;
+    const scaledHeight = this.scale * this.height;
+
+    // create paths for recoloring borders
+    this.topBorder = new Path2D();
+    this.topBorder.moveTo(0, 0);
+    this.topBorder.lineTo(padding, padding);
+    this.topBorder.lineTo(scaledWidth + padding, padding);
+    this.topBorder.lineTo(scaledWidth + 2 * padding, 0);
+    this.topBorder.closePath();
+
+    this.bottomBorder = new Path2D();
+    this.bottomBorder.moveTo(0, scaledHeight + 2 * padding);
+    this.bottomBorder.lineTo(padding, scaledHeight + padding);
+    this.bottomBorder.lineTo(scaledWidth + padding, scaledHeight + padding);
+    this.bottomBorder.lineTo(scaledWidth + 2 * padding, scaledHeight + 2 * padding);
+    this.bottomBorder.closePath();
+
+    this.leftBorder = new Path2D();
+    this.leftBorder.moveTo(0, 0);
+    this.leftBorder.lineTo(0, scaledHeight + 2 * padding);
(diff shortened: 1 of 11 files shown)


browser/static/js/cursor.js
```diff
@@ -0,0 +1,220 @@
...
+  }
+
+  set segArray(newSegArray) {
+    this._segArray = newSegArray;
+    if (this.inRange()) {
+      this.label = Math.abs(this.segArray[this.imgY][this.imgX]);
+    } else {
+      this.label = 0;
+    }
+
```
willgraf, line 44:
```suggestion
```

browser/static/js/cursor.js
```diff
@@ -0,0 +1,220 @@
...
+  }
+
+  // check if the mouse position in canvas matches to a displayed part of image
+  inRange() {
+    if (this.canvasPosX >= 0 && this.canvasPosX < this.width * this.scale &&
+        this.canvasPosY >= 0 && this.canvasPosY < this.height * this.scale) {
+      return true;
+    } else {
+      return false;
```
willgraf, line 53: Just return the boolean condition instead of checking it.

```suggestion
    return (
      this.canvasPosX >= 0 && this.canvasPosX < this.width * this.scale &&
      this.canvasPosY >= 0 && this.canvasPosY < this.height * this.scale
    );
```

browser/static/js/cursor.js
```diff
@@ -0,0 +1,220 @@
...
+      ctx.fillStyle = 'white';
+    } else {
+      ctx.fillStyle = 'black';
+    }
+    ctx.fill(this.bottomBorder);
+
+    ctx.restore();
+  }
+
+
```
willgraf, line 171:
```suggestion
```

browser/static/js/cursor.js
```diff
@@ -0,0 +1,220 @@
...
+    if (tempPanY >= 0 && tempPanY + this.sHeight < this.height) {
+      this.sy = tempPanY;
+    } else {
+      tempPanY = Math.max(0, tempPanY);
+      this.sy = Math.min(this.height - this.sHeight, tempPanY);
+    }
+  }
+
+  changeZoom(dZoom, canvasPosX, canvasPosY) {
+    const newZoom = this.zoom - 10*dZoom;
```
willgraf, line 194:
```suggestion
    const newZoom = this.zoom - 10 * dZoom;
```

browser/static/js/cursor.js
```diff
@@ -0,0 +1,220 @@
...
+      this.label = 0;
+    }
+  }
+
+}
+
+// handle updating zooming and panning attributes
+class CanvasView {
+  constructor(width, height) {
+
```
willgraf, line 81:
```suggestion
```

browser/static/js/cursor.js
```diff
@@ -0,0 +1,220 @@
...
+    this.zoom = 100;
+    this.zoomLimit = 100;
+
+    this.scale = 1;
+
+    this.topBorder = new Path2D();
+    this.bottomBorder = new Path2D();
+    this.rightBorder = new Path2D();
+    this.leftBorder = new Path2D();
+
```
willgraf, line 97:
```suggestion
```

browser/static/js/cursor.js
```diff
@@ -0,0 +1,220 @@
...
+      const propX = canvasPosX / scaledWidth;
+      const propY = canvasPosY / scaledHeight;
+      const dx = propX * (newWidth - oldWidth);
+      const dy = propY * (newHeight - oldHeight);
+      this.pan(dx, dy);
+    }
+  }
+
+
+}
```
willgraf, line 220:
```suggestion
}

```

browser/static/js/cursor.js
```diff
@@ -0,0 +1,220 @@
...
+    this.pressed = false;
+    this.trace = [];
+
+    // mouse coords
+    // mouse position on canvas, no adjustment for padding
+    this.rawX = 0;
+    this.rawY = 0;
+    // adjusted for padding
+    this.canvasPosX = -5;
+    this.canvasPosY = -5;
```
willgraf, line 20: This should probably be an input instead of hardcoded, since we use the `padding` variable in main.

browser/static/js/main_zstack.js
```diff
@@ -412,14 +412,16 @@ class Mode {
...
+      "trace": JSON.stringify(cursor.trace), // stringify array so it doesn't get messed up
       "target_value": brush.target, // value that we're overwriting
       "brush_value": brush.value, // we don't update caliban with edit_value, etc each time they change
       "brush_size": brush.size, // so we need to pass them in as args
       "erase": (brush.erase && !brush.conv),
       "frame": current_frame
     });
-    mouse_trace = [];
+    // TODO: cursor.clearTrace ?
+    cursor.trace = [];
```
willgraf, line 424: Yes!

browser/static/js/main_zstack.js
```diff
@@ -993,11 +905,13 @@ function load_file(file) {
...
-      sy = 0;
-      swidth = rawWidth = rawDimensions[0];
-      sheight = rawHeight = rawDimensions[1];
+      rawWidth = rawDimensions[0];
+      rawHeight = rawDimensions[1];
 
+      viewer = new CanvasView(width=rawWidth, height=rawHeight);
+
+      // TODO: does this need to be in load_file?
+      // we could probably call it right after load_file to decouple them better
```
willgraf, line 914: It doesn't need to be in load_file, but because it is reliant on the return value of the HTTP request (as are ALL of the objects we have).

One way we could try to decouple them is by passing a callback function through `load_file(file, cb)` and defining that callback function inside `startCaliban`. that way we handle all the object creation where we get the raw dimensions, but still have it defined inside our entrypoint.

browser/static/js/main_zstack.js
```diff
@@ -1056,49 +970,21 @@ function setCanvasDimensions(rawDims) {
...
-  scale = Math.min(scaleX, scaleY);
+  const scale = Math.min(scaleX, scaleY);
   // dimensions need to maintain aspect ratio for drawing purposes
   dimensions = [scale * rawDims[0], scale * rawDims[1]];
 
-  zoom = 100;
-  zoomLimit = 100;
+  viewer.zoom = 100;
+  viewer.scale = scale;
+  viewer.setBorders(padding);
```
willgraf, line 983: Could all this rescale logic be encapsulated inside the `viewer`?

browser/static/js/main_zstack.js
```diff
@@ -889,65 +827,38 @@ function render_edit_image(ctx) {
...
-  }
-  ctx.fill(bottomBorder);
-
-  ctx.restore();
 }
 
 function render_image_display() {
   let ctx = $('#canvas').get(0).getContext('2d');
   ctx.imageSmoothingEnabled = false;
+  // TODO: is there a corresponding ctx.restore to match this ctx.save?
```
willgraf, line 861: I'm not sure I understand what we want to achieve here.

browser/static/js/main_zstack.js
```diff
@@ -1260,6 +1141,7 @@ function action(action, info, frame = current_frame) {
         // array of arrays, contains annotation data for frame
         if (payload.imgs.hasOwnProperty('seg_arr')) {
           seg_array = payload.imgs.seg_arr;
+          cursor.segArray = seg_array;
```
willgraf, line 1144: does the global `seg_array` still need to exist? Can we just directly assign `cursor.segArray = payload.imgs.seg_arr;`?

PR #198, opened 2020-08-14
Merge Cursor and View objects.

willgraf:
The Cursor and Viewer classes share a lot of attributes and are often updated at the same time. I think that merging them together into a `CalibanCanvas` class makes using them/it a bit easier.

I also made some general quality-of-life improvements:
- I've added some ESLint files that support linting in most editors. 
- I've changed `load_file` to pass a callback function, which we can define in `startCaliban`, which moves the initialization of some of our objects close to where we define the inputs.
- I've tried to use vanilla javascript where possible, instead of relying on jquery.
- I may have gone a little crazy linting some of the `main_zstack.js`.


browser/static/js/canvas.js
```diff
@@ -206,15 +191,31 @@ class CanvasView {
...
+      this.scaledWidth,
+      this.scaledHeight
+    );
+  }
 
-}
\ No newline at end of file
+  updateCursorPosition(x, y, padding) {
+    this.updatePos(x, y, padding, this);
+  }
```
geneva-miller, line 216: could we just rename `updatePos` to `updateCursorPosition` and get rid of this method?

willgraf: yes! good catch.

PR #201, opened 2020-08-31
add a white outline to highlighted pixel-editing label

geneva-miller:
Add a white outline to highlighted label in non-rgb pixel-editing mode to help distinguish labels from each other. Highlighting already recolored current label to red, but the semi-translucency of the label could still make it a little difficult to see exactly where the boundaries were. This change makes the highlighting behavior more similar to desktop version (label is outlined but not changed to red) and to how the conversion brush labels are shown in non-rgb pixel-editing mode.

Example:
image

### DIFF
--- a/browser/static/js/adjust.js
+++ b/browser/static/js/adjust.js
@@ -285,7 +285,7 @@ class ImageAdjuster {
     this.compositedImg.src = this.canvas.toDataURL();
   }
 
-  // apply outlines, transparent highlighting
+  // apply white (and sometimes red) opaque outlines around cells, if needed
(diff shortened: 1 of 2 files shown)


browser/static/js/adjust.js
```diff
@@ -304,6 +304,10 @@ class ImageAdjuster {
       singleOutline = true;
       o1 = brush.value;
     }
+    if (editMode && currentHighlight && !brush.conv) {
+      singleOutline = true;
+      o1 = brush.value;
+    }
 
     this.postCompositeLabelMod(
```
willgraf, line 312: Can you just add a comment here about what this is doing? The conditional is not obvious to me.

PR #202, opened 2020-08-31
Python backend refactor

tddough98:
This refactor breaks down caliban.py into three modules: files.py, views.py and reviews.py.

The first module files.py includes one class BaseFile that stores the numpy arrays and metadata needed to view an image in Caliban. This class consolidates the metadata for .npz (`cell_info` and `cell_ids`) and .trk (`tracks`) files into `cell_info` and `cell_ids`. The `tracks` attribute is now an alias for `cell_info[0]` and the lineages from the .trk overwrite `cell_info` to keep the class backwards compatible.

The second module views.py includes one class BaseView that has a single BaseFile attributes, as well as attributes to store which part of the image we are viewing. The class separates the methods and attributes needed to view a BaseFile from those needed to an edit a BaseFile, which remain in reviews.py.

The third model reviews.py includes three classes, all originally from caliban.py: BaseReview, ZStackReview, and TrackReview. The BaseReview class extends the BaseView class, adding an output_bucket attribute and methods to edit its BaseFile attribute.

This PR also includes tests for all three of these modules, although not every method is tested yet. The tests all use new BaseFile fixtures in conftest.py that construct some simple numpy arrays (like an array with no label, one label on the whole screen, one label on half the screen, two labels on each half, checkboard labels, etc.).

Below is a simple UML diagram of the classes in these classes.
Caliban objects

Other small changes include:

- AJAX requests in the caliban JavaScript now start their URLs with `document.location.origin`
- action_change_channel and action_change_feature can't be set to channels_max or features_max anymore (which would have been an indexing error)


tddough98:
I've made the following changes after talking with Will:

- remove feedback.js
- rename classes (BaseFile -> CalibanFile, BaseView -> View, Review -> Edit)
- use a factory pattern to instantiate Edit objects
- alias raw and annotated file attributes in the View class (to remove self.file... pattern)
- combine file.py & review.py into caliban.py
- write docstrings for some methods in caliban.py

browser/blueprints.py
```diff
@@ -334,3 +337,11 @@ def shortcut(filename):
...
         filename=filename,
         settings=settings)
+
+
+# Factory for Edit objects
+def get_edit(file_, output_bucket, rgb):
+    if is_npz_file(file_.filename):
+        return ZStackEdit(file_, output_bucket, rgb)
+    elif is_trk_file(file_.filename):
+        return TrackEdit(file_, output_bucket)  # don't use RGB mode with track files
```
willgraf, line 347: Nice! May as well put the comment as a docstring:

```suggestion
def get_edit(file_, output_bucket, rgb):
    """Factory for Edit objects"""
    if is_npz_file(file_.filename):
        return ZStackEdit(file_, output_bucket, rgb)
    elif is_trk_file(file_.filename):
        return TrackEdit(file_, output_bucket)  # don't use RGB mode with track files
```

Also, what should this function do if the file is neither a track of zstack file?

browser/caliban.py
```diff
@@ -105,13 +90,39 @@ def add_outlines(self, frame):
...
         return frame
 
     def get_frame(self, frame, raw):
+        """
+        Returns the ith frame to display on the front end.
+        Also changes the current frame of this View.
+
+        Args:
+            frame (int):
+            raw (bool): when
```
willgraf, line 115: Missing some details.

browser/conftest.py
```diff
@@ -40,3 +42,87 @@ def _db(app):
...
+
+
+@pytest.fixture(params=TEST_FRAMES, ids=TEST_NAMES)
+def zstack_file(mocker, request):
+    def load(self):
+        data = {'raw': np.zeros((FRAMES, HEIGHT, WIDTH, CHANNELS))}
+        data['annotated'] = repeat_frame(repeat_feature(request.param, FEATURES), FRAMES)
+        return data
+    mocker.patch('files.CalibanFile.load', load)
+    return CalibanFile('filename.npz', 'bucket', 'path')
```
willgraf, line 102: It might be useful to try testing a 2D image (`rgb` mode).

browser/static/js/main_zstack.js
```diff
@@ -752,7 +752,7 @@ const _calculateMaxHeight = () => {
 function upload_file(cb) {
   $.ajax({
     type: 'POST',
-    url: `upload_file/${project_id}`,
+    url: document.location.origin + `/upload_file/${project_id}`,
```
willgraf, line 755:
```suggestion
    url: `${document.location.origin}/upload_file/${project_id}`,
```

Can you test if this works?

Also, can you update `track.js` to use this as well?

browser/files_test.py
```diff
@@ -0,0 +1,22 @@
+"""Test for File classes"""
+
+import pytest
+import numpy as np
+
+from files import CalibanFile
+
+
+def test_init(file_):
```
willgraf, line 9: Consider adding a test for the `create_cell_info` method, though it may overlap with the current tests.

willgraf (changes requested):

browser/blueprints.py
tddough98: As the factory takes a CalibanFile object as an argument, the CalibanFile object would have already thrown an error if it were not passed a track or zstack file. Still, a BaseEdit could be an appropriate default case.

tddough98:
I've made all the requested changes, as well as two extra changes:

- trials is no longer an attribute in CalibanFile, and instead is just kept as a local variable in `__init__` while we set the raw, annotated, and tracks
- I've realized several of my tests in caliban_tests.py assume that each label is present in every frame, which is true for the current test suite, but not in general. I've added comments & TODOs where I make this assumption in the relevant tests

browser/conftest.py
tddough98: I added a single frame test image, which I think is what we're looking for. This exercises the dimension expansion code (starting with `if self.raw.ndim == 3:`) in the CalibanFile `__init__` that was previously grouped with the RGB logic.

willgraf (approved):
Awesome!

tddough98:
These changes address the issues that Geneva found with editing .trk files, including:
- tracks in main_track.js now loads payload.tracks[0] to account for .trk files now using cell_info to store metadata instead of tracks; info like frames, parent, daughters, etc. now displays properly
- action_new_track now updates cell_ids & does nothing when creating a new track on the first frame of a track (previously action_new_track did nothing only on frame 0)
- action_new_track is now tested

browser/static/js/main_track.js
```diff
@@ -788,7 +788,7 @@ function handle_mousemove(evt) {
 function handle_mouseup(evt) {
   mousedown = false;
   if (edit_mode) {
-    //send click&drag coordinates to caliban.py to update annotations
+    //send click&drag coordinates to reviews.py to update annotations
```
geneva-miller, line 791: This got changed back to `caliban.py` from `reviews.py`
```suggestion
    //send click&drag coordinates to caliban.py to update annotations
```

geneva-miller (commented):
This looks great! Both trk and npz modes work as expected in the browser.

browser/README.md
```diff
@@ -55,7 +55,7 @@ Functions depend on Python libraries -- including NumPy, Matplotlib, and scikit-
...
 
-An SQLite database has been added for local use of Caliban. SQLite is embedded into the end program, so whenever the tool is started, a unique id is created for the user and the TrackReview/ZStackReview object is stored locally in caliban.db. Whenever the user makes a change to the object, the application will update the state of the object. After the user submits the final datafile to the S3 bucket, the corresponding id and object will be deleted from the database file. The caliban.db file can be created simply by running application.py.
+An SQLite database has been added for local use of Caliban. SQLite is embedded into the end program, so whenever the tool is started, a unique id is created for the user and the TrackEdit/ZStackEdit object is stored locally in caliban.db. Whenever the user makes a change to the object, the application will update the state of the object. After the user submits the final datafile to the S3 bucket, the corresponding id and object will be deleted from the database file. The caliban.db file can be created simply by running application.py.
```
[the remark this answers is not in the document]
tddough98:
```suggestion
Caliban can also be run locally using a SQLite database (this is the default behavior). When we start using Caliban, Caliban creates a TrackEdit (for .trk files) or ZStackEdit (for .npz files) object, gives it a unique ID, and stores it locally in caliban.db. Whenever we change an Edit object, Caliban updates the object in the database. After we submit the file and our changes to the S3 bucket, Caliban deletes the Edit object from the database, leaving behind the unique ID and session metadata. Running application.py creates the database if it does not already exist.
```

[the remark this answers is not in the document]
geneva-miller: Looks good!

geneva-miller (approved):

PR #206, opened 2020-09-17
Additional browser js refactoring with CanvasState

### LINKED ISSUE #158: handle_draw sends duplicate mouse trace coordinates in payload.

### LINKED ISSUE #203: fully replacing label with conv brush doesn't clear it from cell_info
This is only an issue in the browser version. If a label is completely erased from a frame using the normal brush, the file metadata is updated to reflect that the label no longer appears in that frame. When the conversion brush functionality was added to browser, the add/erase logic was not updated to reflect that the label being drawn over can be fully removed from the frame (since previously, that label could only be 0 and never had metadata). 

`handle_draw` should be fixed to update file metadata if the conversion brush is used to totally overwrite a label.

geneva-miller:
## What
* Further refactoring of javascript with focus on CanvasState (make `padding` an object attribute instead of global variable, add methods to handle trace when drawing)
* Updates `action_handle_draw` functionality in `caliban.py` backend

## Why
* Reduce globally scoped variables (another step towards improved modularity of code)
* Abstraction of mouse trace behavior into CanvasState methods leads to easier implementation of drawing with mouse events
* Close #158 (improves efficiency of drawing action)
* Close #203 (prevents conversion brush cell info bug)

### DIFF
--- a/browser/static/js/canvas.js
+++ b/browser/static/js/canvas.js
@@ -46,6 +46,18 @@ class CanvasState {
     this.isSpacedown = false;
   }
 
+  reset(scale, padding) {
+    this.scale = scale;
+    this.zoom = 100;
+
+    // set viewing coords back to show full image
+    this.sx = 0;
+    this.sy = 0;
+    this.sWidth = this.width;
+    this.sHeight = this.height;
+    this.setBorders(padding);
+  }
+
   get segArray() {
     return this._segArray;
   }
--- a/browser/static/js/canvas.js
+++ b/browser/static/js/canvas.js
@@ -38,7 +38,7 @@ class CanvasState {
     // store as part of object to be able to get current label
     this._segArray = null;
 
-    this.scale = scale;
+    this._scale = scale;
 
     this.topBorder = new Path2D();
     this.bottomBorder = new Path2D();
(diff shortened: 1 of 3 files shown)


browser/static/js/canvas.js
```diff
@@ -67,9 +88,54 @@ class CanvasState {
...
+  inTrace(newCoord) {
+    let duplicate = false;
+    for (var i=0; i < this._trace.length; i++) {
+      let coord = this._trace[i];
+      if (coord[0] === newCoord[0] && coord[1] === newCoord[1]) {
+        duplicate = true;
+        break;
+      }
+    }
+    return duplicate;
```
willgraf, line 138: O(n) lookup speed on every mouse move seems inefficient. Let's brainstorm some other approaches. Could be the Coordinate object, which we can lookup in a map, or maybe we can set our own 2d boolean array and only add coords that are still False in the array?

willgraf (changes requested):
Great start, I love the `paddedWidth` getters!

In addition to the comments, I think this PR should also update the track functionality to use the new methods as well.

browser/static/js/canvas.js
geneva-miller: I'm not sure how to implement the Coordinate object + map, but I'm intrigued by the 2d boolean array idea

PR #209, opened 2020-10-10
Performance refactor

tddough98:
## What
This PR provides a comprehensive rework of Caliban's data models, focused on finer grained updates to the project's data. Previously, all of a project's logic and data lived in the ZStackEdit or TrackEdit class, and we pickled the whole class in and out of the database for each action.

The data models now separate the project from its frames and its metadata. We use SQLAlchemy ORM to manage the data, and the Edit classes now take a row from the Project table, perform an action, and update the changed parts of the Project. A UML diagram of the SQLAlchemy ORM classes is show below.

RefactoredDataModels (1)

This PR also adds Flask-MonitoringDashboard (FMD) to caliban, which automatically monitors the performance of all endpoints and provides profiling and performance comparisons between endpoints, versions, blocks of time, and IP addresses. This PR includes logic to separately compare the performance of each caliban action, which appears in the "Per-User" comparison tab.

## Why
With finer data models, actions that change a single frame can update only that frame, rather than pickling and rewriting the entire project. We see dramatically improved performance for most actions, especially single frame actions, as their database operations become lightweight. Even for actions that edit every frame in the image stack, we gain a performance boost as no longer need to read and write the raw or RGB frames with every action.


tddough98:
The Flask Monitoring Dashboard offers constant monitoring of all of the endpoints of a flask application. An example overview page for Caliban appears below:
Screen Shot 2020-10-12 at 11 16 06 AM

For each endpoint, FMD offers 4 levels of monitoring from
- 0 - Disabled
- 1 - Performance
- 2 - Outliers
- 3 - Profiler + Outliers
We monitor all the endpoints at level 3 right now, except for caliban.form and caliban.health, which are disabled.

For each endpoint, we can view
- Hourly API Utilization
- User-Focused Multi-Version Performance
- IP-Focused Multi-Version Performance
- Per-Version Performance
- Per-User Performance
- Profiler
- Grouped Profiler
- Outliers
- Status Code Distribution

The Hourly API Utilization, User-Focused Multi-Version Performance, and IP_Focused Multi-Version Performance all show heatmaps, showing when an endpoint is requested, the median latency by user and version, and the median latency by IP-address and version, respectively. The Per-Version Performance and Per-User Performance show box and whiskers plots of endpoint latency by version or user. The Profiler shows line-by-line profiling for each request, and the Grouped Profiler combines the line profiling for all requests into a sunburst plot. Outliers shows the extra info collected for outlier requests, and the Status Code Distribution shows the returned HTTP status codes (like 200 or 404).

The User-Focused & Per-User Performance uses FMD's group_by configuration, and expects a function that returns a user ID. As we direct many action types (like `handle_draw` or `change_channel`) to the same caliban.action endpoint, we instead group_by the action type, as defined here. Given that FMD expects groupings by user ID, the action groupings are in the User-Focused/Per-User tabs, resulting in plots like this:
Screen Shot 2020-10-12 at 11 30 21 AM

I've also deployed an [example EB instance](tdougherty-test.us-east-2.elasticbeanstalk.com) for the performance improvements to play with and test the improvements.

browser/models.py
```diff
@@ -3,83 +3,664 @@
...
+    rgb_frames = db.relationship('RGBFrame', backref='project')
+    label_frames = db.relationship('LabelFrame', backref='project')
+    # Includes underscore to avoid conflict with metadata (MetaData) attribute from SQLAlchemy
+    metadata_ = db.relationship('Metadata', backref='project', uselist=False)
+
+    def __init__(self, filename, input_bucket, output_bucket, path,
+                 rgb=False, raw_key=None, annotated_key=None):
+        init_start = timeit.default_timer()
+        if raw_key is None:
+            raw_key = 'raw'
```
willgraf, line 60: Why not just use as default `raw_key='raw'`?

browser/models.py
```diff
@@ -3,83 +3,664 @@
...
+        current_app.logger.debug('Created RGB frames for %s in %ss.',
+                                 filename, timeit.default_timer() - start)
+
+        start = timeit.default_timer()
+        self.label_frames = [LabelFrame(self.id, i, frame)
+                             for i, frame in enumerate(annotated)]
+        current_app.logger.debug('Created label frames for %s in %ss.',
+                                 filename, timeit.default_timer() - start)
+        # Log total time in constructor
+        current_app.logger.debug('Total time in project __init__ for %s: %s s.',
```
willgraf, line 101: Can we change the log text? Maybe something like `"Project for "%s" initialized in %s s."`?

willgraf (changes requested):
I haven't read through all the test files, but thought I'd post some suggestions I have so far.

browser/blueprints.py
```diff
@@ -90,89 +77,64 @@ def upload_file(project_id):
...
-    # Obtain raw, mask, and edit mode frames
-    img = state.get_frame(frame, raw=False)
-    raw = state.get_frame(frame, raw=True)
-
-    # Obtain color map of the cells
-    edit_arr = state.get_array(frame)
-
+    # Change the frame
+    project.metadata_.frame = frame
+    db.session.commit()
```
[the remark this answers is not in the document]
tddough98: I made an update method in the Project table that contains the `db.session.commit()` statement. I also replaced the `db.session.commit()` line with `project.update()` in the `persist_pickles` method (which I also renamed to `commit_changes`).

Because I made `update` an instance method, I changed `finish_project` from a static method to an instance method & renamed it to `finish` (so we have `project.finish()` instead of `Project.finish_project(project)`). I also renamed `get_project` to `get` and `create_project` to `create`, but they are both still static methods as they interact with a the Project table, not an instance.

browser/caliban.py
```diff
@@ -539,210 +419,258 @@ def action_threshold(self, y1, x1, y2, x2, frame, label):
...
+        Copy the PickleType columns that have changed to persist them in the database.
+        """
+        if self.info_changed:
+            self.metadata.update()
+        if self.multi_changed:
+            for label_frame in self.project.label_frames:
+                label_frame.update()
+        elif self.y_changed:
+            self.project.label_frames[self.frame_id].update()
+        db.session.commit()
```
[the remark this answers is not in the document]
tddough98: Now replaced with a call to `self.project.update()`.

browser/models.py
tddough98: I think I just did the same thing for both raw and annotated, but it was only needed for annotated. Fixed to put raw_key='raw' in the function signature instead.

browser/models.py
tddough98: Changed to `'Initialized project for %s in %s s.'` to be most similar to the other log statements in Project `__init__`

geneva-miller:
One thing I'm curious about so far: it looks like for single-frame actions, `frame` has been removed as an action kwarg and the action just takes place in the current frame. Previously, including `frame` as an action kwarg meant that a caliban user could select a label in a frame, change frames, and still have their action be performed on the frame they had selected. It's an edge case (I think most of the time, users shouldn't be accidentally changing frames before confirming an action), and we don't strictly need to keep that behavior. 

Should we try and add that behavior back in? Or should we add some logic to the frontend to prevent that type of thing from happening? As it currently stands I think there is a mismatch between what the frontend indicates (eg, "Delete label 5 in frame 0?" while in frame 1) and what will actually happen if the user changes frames before confirming an action (eg, label 5 gets deleted from frame 1, not frame 0).

tddough98:
> One thing I'm curious about so far: it looks like for single-frame actions, `frame` has been removed as an action kwarg and the action just takes place in the current frame. Previously, including `frame` as an action kwarg meant that a caliban user could select a label in a frame, change frames, and still have their action be performed on the frame they had selected. It's an edge case (I think most of the time, users shouldn't be accidentally changing frames before confirming an action), and we don't strictly need to keep that behavior.
> 
> Should we try and add that behavior back in? Or should we add some logic to the frontend to prevent that type of thing from happening? As it currently stands I think there is a mismatch between what the frontend indicates (eg, "Delete label 5 in frame 0?" while in frame 1) and what will actually happen if the user changes frames before confirming an action (eg, label 5 gets deleted from frame 1, not frame 0).

Ah I didn't realize that this behavior was intended. I thought that the frame sent by the frontend and the frame recorded in the backend database would always be in sync, so I started using just the backend frame to simplify some things.

I think the best way to restore this behavior would be to add a frame argument to the Edit constructor to be `ZStackEdit(project, frame)` instead of `ZStackEdit(project)`, as this extra frame info is specific to the action we're doing, rather than the larger project. Previously, each action took a frame argument, but this would be clumsy now as each action would have to access the right frame from the database, instead of using the frame attribute of the Edit class.

On the other hand, if we don't need this behavior, I agree that mismatch on the frontend is an issue. If a user selects label 5 on frame 0, then switches to frame 1, we should switch the info message to say ""Delete label 5 in frame 1?" instead. If we change prompts to use `current_frame` instead of `this.info.frame`, then we'll have a prompt that reflects the correct frame when we begin the action, but if we change the frame before confirming the action, then there will still be a mismatch. As we store the prompt as a string, I'm not sure the best way to handle this mismatch... I'm thinking that we could refactor the action or prompt into a class, but this needs more thought & design. Such a class could collect all the actions in the same place & make it easy to reference or modify the actions

geneva-miller:
I think actions like `watershed` or `flood_fill` are the particular reason why frame was also passed as an argument. If the click location doesn't correspond to the same cell label in a different frame, those actions might have unexpected behavior. From that perspective I think I would you to include frame as an optional arg to `ZStackEdit` for now. (I would also like to refactor actions and prompts into their own class(es) in the frontend but I think that's out of scope for this PR.)

browser/caliban.py
```diff
@@ -539,210 +417,259 @@ def action_threshold(self, y1, x1, y2, x2, frame, label):
...
 
 class TrackEdit(BaseEdit):
-    def __init__(self, file_, output_bucket):
-        super(TrackEdit, self).__init__(file_, output_bucket)
+    def __init__(self, project):
+        super(TrackEdit, self).__init__(project)
 
-        self.scale_factor = 2
+        # TODO: do we need this?
+        # self.scale_factor = 2
```
geneva-miller, line 647: we do need this for now but it will be removed soon (once the frontend makes use of `CanvasState` for trk files--in progress)

geneva-miller (commented):
Looks good to me

tddough98:
> I think actions like `watershed` or `flood_fill` are the particular reason why frame was also passed as an argument. If the click location doesn't correspond to the same cell label in a different frame, those actions might have unexpected behavior. From that perspective I think I would like you to include frame as an optional arg to `ZStackEdit` for now. (I would also like to refactor actions and prompts into their own class(es) in the frontend but I think that's out of scope for this PR.)

What we instead cleared the info/action/prompt when switching frames? Then a user will always submit info that's consistent with the frame that they're looking at, and there will never be a misleading prompt/unexpected result. This would be straightforward to implement, as we would only need to call clear() whenever we call fetch_and_render_frame().

This option would completely remove the ability to change frames during an action, but it seems reasonable to have users view the frame that they're editing and the same range of results are possible without this behavior.

geneva-miller:
> What we instead cleared the info/action/prompt when switching frames?
I like this idea, but I'll note that we should not clear `replace` actions--replacing labels in different frames is a common use case. Off the top of my head, the rest of the actions should be okay to clear upon switching frames.

tddough98:
I've made the following small/cosmetic changes:
- renamed Metadata table to State
- added a test project to conftest.py that uses RGB mode
- added a unit test for add_outlines in imgutils.py
- renamed numFrames, numChannels & numFeatures to use snake_case

I've also addressed the issue that Geneva raised where Caliban previously performed an action where a label was selected, but not Caliban performs the action on the current frame. I've kept the new behavior to act on the current frame, as then changes are more apparent & intuitive to the user. 

I've worked around the inconsistent prompts from Caliban (like "Delete label 1 from frame 0?" when we actually delete from a different frame) by calling mode.clear() when changing frames _if there is an action selected_. This allows the user to select labels & change frames without losing their selection, but once they initiate an action, they must complete the action on the same frame.

This solution isn't 100% perfect, and in particular, I can see issues arising from watershed. Watershed uses the selection position as its seed, so if we select a position and change frames, the label at that position may have changed. At worst, watershed internally throws an error and the user sees no change on their end. I think this is something to pay closer attention to when we better refactor actions on the front-end, but further work here is out of scope for now.

browser/caliban.py
```diff
@@ -539,210 +417,259 @@ def action_threshold(self, y1, x1, y2, x2, frame, label):
...
```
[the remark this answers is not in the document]
geneva-miller: Looks great! Thanks for implementing the current-frame workaround. Although watershed bugs are still possible, they seem unlikely and pretty minor, so I'm fine with it for the time being. Trk files still need `scale_factor` but once the trk frontend is switched to make use of `CanvasState`, I don't think any of the objects will need to store a `scale_factor` attribute.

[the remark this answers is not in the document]
willgraf: Just for reference, why does tracking need `scale_factor = 2`?

browser/static/js/main_zstack.js
```diff
@@ -382,12 +382,10 @@ class Mode {
...
         action('predict_single', { frame: current_frame });
       } else if (this.action === 'replace') {
-        if (this.info.label_1 !== this.info.label_2 &&
-            this.info.frame_1 === this.info.frame_2) {
+        if (this.info.label_1 !== this.info.label_2) {
           action('replace_single', {
             label_1: this.info.label_1,
-            label_2: this.info.label_2,
-            frame: this.info.frame_1
+            label_2: this.info.label_2
```
willgraf, line 388: We no longer include `frame` in the action payload because we are replacing it in the "active" frame. Was `frame_1` always considered the "active" frame as well? I just want to make sure we aren't changing any behavior here.

browser/caliban.py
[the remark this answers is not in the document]
geneva-miller: The scaling factor of 2 specifically is historically what was used for trk files. Having a `scale_factor` attribute at all is from pre-`CanvasState` frontend code. The scale factor gets sent in the first trk payload to the frontend as part of initializing the canvas with the correct size. The scale factor is subsequently used as part of image processing (frontend) and mapping the cursor location to the correct point in the image (frontend and some places in backend). The changes that `CanvasState`/zooming bring remove the need for `scale_factor` to be stored in the backend--it becomes implicitly 1, nothing is rescaled. The scale factor in the frontend becomes "whatever makes the image fit". So, tracking only needs `scale_factor = 2` for now until it can be refactored out, which is in progress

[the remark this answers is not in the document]
willgraf: OK let's uncomment it here and add a TODO comment to remove it during the next refactor.

tddough98:
We now set scale_factor while creating a row in the Projects table, instead of setting the it in the Edit class constructor. We create Edit objects before each action, but not when the project is initialized, so changes to the project made by Edit would not be in effect while loading the project.

With this change, this PR should be ready to merge. I'll post the undo/redo PR soon, and then once that PR is in, we should be set to archive the existing caliban database and deploy a new one!

browser/application.py
```diff
@@ -80,6 +80,18 @@ def create_app(**config_overrides):
 
     compress.init_app(app)
 
+    # For flask monitoring dashboard
+    dashboard.config.init_from(file='fmd_config.cfg')
```
willgraf, line 84: Does this config file exist?

willgraf (commented):
This looks great, a quick question about the configuration file for the monitoring dashboard but otherwise it looks awesome.

I'd like to merge this in before the Undo button PR (just for history reasons).

browser/application.py
tddough98: This config file is in my local copy of the repo, but not tracked, similar to .env. The config file contains database credentials and dashboard credentials here are the non-sensitive contents of the config file I've been using:

```
[dashboard]
VERSION=refactor
GIT=../.git/
CUSTOM_LINK=dashboard
MONITOR_LEVEL=3
OUTLIER_DETECTION_CONSTANT=2.5
SAMPLING_PERIOD=20
ENABLE_LOGGING=True

[visualization]
TIMEZONE=US/Pacific
```

The sensitive fields (with sensitive values removed) are:
```
[authentication]
USERNAME=
PASSWORD=
SECURITY_TOKEN=

[database]
TABLE_PREFIX=fmd
DATABASE=
```

willgraf: Is it a problem if the file does not exist? Should include a default one without the sensitive fields?

tddough98: I've added the an example config file with all the info above, except for a blank version name

willgraf: I think we should refactor the filename to [be an environment variable](https://flask-monitoringdashboard.readthedocs.io/en/v1.10.5/configuration.html), and maybe we can use the above example file with the defaults filled in?

`dashboard.config.init_from(envvar='DASHBOARD_CONFIG')`

tddough98: I've updated .env.example to include `DASHBOARD_CONFIG=fmd_config.cfg.example` and added the above code line to application.py. It looks like things are working out of the box. With the USERNAME and PASSWORD fields blank, the dashboard defaults to admin/admin credentials, and with DATABASE blank, it defaults to a SQLite database flask_monitoringdashboard.db.

willgraf: Instead of using the `env_var` arg, what do you think about adding `DASHBOARD_CONFIG` to `config.py`?

Also, do we want to always enable it or can we toggle whether to use it? Maybe if `DASHBOARD_CONFIG` is empty we don't need to initialize the dashboard?

tddough98: I like that behavior. As an empty string is falsy, I wrapped the dashboard initialization in application.py in an `if config.DASHBOARD_CONFIG` block

willgraf: beautiful!

willgraf: Last thing I can think of - Should we get rid of the `.example` extension on the configuration file?

geneva-miller (approved):
Looks great to me. I'll defer to Will for any remaining small changes to the configuration file

browser/application.py
tddough98: I buy remove the .example extension for the dashboard config. I included it because we also included it for .env.example, but I suppose that .env.example is nearly empty while the dashboard config is nearly complete, other than the authentication info. I'm going to remove the .example from the dashboard config file and add a comment inside it to highlight the default behavior of the empty fields.

tddough98: Hmmmm, I just realized that if we remove the `.example` from the dashboard config file, then git will be aware of changes to it and prompt us to track the changes, which would often include sensitive info like credentials. By explicitly including the `.example` extension, the user can more easily create an untracked dashboard config file within their repo by removing `.example`.

willgraf: I'm convinced.

willgraf (approved):
Great work! Lot's of great improvements for improving/monitoring the performance of Caliban.

PR #210, opened 2020-10-21
no longer require label 1 and validate all tracks

jowch:
## What
* `TrackReview` scans all tracks for required keys
* `TrackReview` remembers all invalid tracks and prints missing keys
* `TrackReview` no longer requires label "1"
* `TrackReview` uses smallest label as default `display_info`

## Why
* Removing or replacing the label "1" during annotation results in a corrupted track file with the existing code. That code specifically checked for the label "1" and validated only the track with that label. This meant that users who are not able to fix the error manually by unzipping the .trk file would have to start over.

### DIFF
--- a/desktop/caliban.py
+++ b/desktop/caliban.py
@@ -1942,15 +1942,23 @@ class TrackReview(CalibanWindow):
         self.tracked = tracked
 
         # if not all of these keys are present, actions are not supported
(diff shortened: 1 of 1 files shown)


jowch:
Sorry! Didn't mean to PR this here! 😅

PR #211, opened 2020-10-28
Add .ebignore file & install gcc during ELB deployment.

willgraf:
## What
* Add `.ebignore` file mostly based on the `.gitignore`.
* Add `01_packages.config` file for configuring ELB to install `gcc` on Amazon Linux.

## Why
* Prevent bloated artifacts and speed up deployment.
* Prevent failure mode where new instances on ELB can fail to install due to a missing `gcc`.


willgraf:
Closing this as it is seemingly unnecessary, though I think the `.ebignore` file is worth keeping as it makes deployments much faster. I'll make a different PR for that.

PR #212, opened 2020-11-03
Undo

tddough98:
## What

We introduce an undo/redo button to the ZStack app for Caliban. We do not implement undo/redo for tracking, only making a few changes to keep the app working. Future work could add undo/redo to tracking by merging main_track.js and main_zstack.js.

The undo/redo button is headed by an action manager, written in `history.js`. The action manager takes and action, completes the action, and maintains groups of actions to undo or redo. These actions can be grouped with "fenceposts" to undo a series of actions in one click, like a series of small `Pans` to be undone as one movement. Each action must implement the `Action` interface, where it must implement `do()`, `undo()`, and `redo()` methods. Each action has access to the data needed to do, undo and redo the action, such as parameter values or the objects it manipulates. Clients of the action manager will create an Action instance, and later will call addAction(action) to record and carry out the action. This design is based on the [command design pattern](https://en.wikipedia.org/wiki/Command_pattern).

Here is the public interface for the action manager:
- `addAction(action)`: does and manages an action
- `addFence()`: places a "fencepost" on top of the undo stack to mark the boundary of an undoable group
- `addFencedAction(action)`: adds fenceposts on both sides of an action to create a group with a single action
- `undo()`: undoes the previous group of actions group
- `redo()`: redoes an undone group of actions

The currently implemented Action classes are:
- in `canvas.js`
  - `Pan`
  - `Zoom`
- in `adjuster.js`
  - `ChangeContrast`
  - `ChangeBrightness`
  - `ResetBrightnessContrast`
  - `ToggleInvert`
- in `zstack_actions.js`
  - `ToggleEdit`
  - `ToggleHighlight`
  - `ChangeFrame`
  - `ChangeFeature`
  - `ChangeChannel`
  - `BackendAction`

Notably, `ChangeFrame`, `ChangeFeature`, and `ChangeChannel` all use a new route `changedisplay`, intended for changes that do not edit the data, but rather change the displayed data. Previously, changing frames was handled by the `get_frame` route, while changing feature and channels were handled by the `action` route. The `load` route now includes image data in its payload to account for the removal of the `get_frame` route. The `action` route now only deals with actions that edit the Project's labeling and has been renamed to `edit`.

We now include two routes `undo` and `redo` which use the stored frames and labels in the backend action history to undo or redo `edit` route calls that change the labeling. These routes are only used to undo/redo `edit` requests, and not `changedisplay` requests, which we instead undo/redo with another `changedisplay` request.

Below are the updated SQLAlchemy data models, now including the action history with the Action table and FrameHistory table.

UndoDataModels (1)

The Projects table is the entry point into the data. Its public interface includes two static methods:
- `Project.get(project_id)`: get a row from the Projects table
- `Project.create(filename, input_bucket, output_bucket, path)`: create a row in the Projects table
  - Caliban still only supports loading files from S3. Future work could develop a file loader interface and pass file loader instance to create instead.

as well as the following instance methods for each row from the Projects table: 
- `update()`: completes the unit of work pattern and commits changes to the database
- `finish()`: clears out BLOB columns from the database (except for the action history)
- `finish_action(action_name)`: finish the current action & save the project state before the next action;
  - only called by `edit` route before calling update()
  - needs to be separate from `update()` so that `changedisplay` can update the project without creating a new action
- `undo()`: restores the project to before a previous label-editing action
- `redo()`: restores the project to after an undone label-editing action
- `get_max_label()`: called by label-editing actions that need to use an unused label

Other changes include:
- use SQLAlchemy's `Mutable` mixin to subclass `numpy.ndarray` and detect changes to the array
  - the `cell_info` and `cell_ids` columns in the `Labels` table still use vanilla dictionaries and must be explicitly copied for SQLAlchemy to detect and persist changes. Future work could use the `Mutable` mixin for these columns as well
- add `session_options={"autoflush": False}` to database so accessing relationship columns (like project.raw_frames) does not flush changes
- enable page scrolling outside when the cursor is not on the canvas
- renaming javascript variables, most prominently `state` --> `canvas`
- fixed scrolling on mouse wheels to adjust brightness (previously only worked on trackpads as shift + scroll turns on horizontal scrolling for wheels)

## Why

The lack of an undo button has been highlighted as one of the most glaring usability issues. Annotators have noted that without an undo, they have had to restart projects instead of using other tools to fix an error. 

Additionally, we can better understand how annotators label/correct our data by capturing the series of changes and actions used within Caliban. This PR does not clear the action history upon submission so that we can build a dataset of Caliban actions/changes.


willgraf:
How does this relate to #209?

tddough98:
I branched off #209 to do this work, so all the performance changes are contained within this PR as well. The few changes made to #209 after branching off for undo/redo (like changing the scale_factor for .trk files) have been merged into this branch as well.

browser/static/js/history.js
```diff
@@ -0,0 +1,135 @@
...
+      }
+    }
+    this.formatButtons();
+  }
+
+  formatButtons() {
+    document.getElementById('undo').disabled = !this.canUndo;
+    document.getElementById('redo').disabled = !this.canRedo;
+  }
+}
```
willgraf, line 135:
```suggestion
}

```

browser/static/js/history.js
```diff
@@ -0,0 +1,135 @@
...
+    }
+    // Undo actions until the end of the group or stack
+    while(true) {
+      action.undo();
+      this.redoStack.push(action);
+      if (this.undoStack.length === 0) break;
+      action = this.undoStack.pop();
+      if (action === 'fencepost') {
+        // Keep fencepost on top of undo to separate new actions from last undo
+        this.undoStack.push('fencepost');
```
willgraf, line 100:
```suggestion
        this.addFence()
```

browser/static/js/canvas.js
```diff
@@ -215,3 +171,86 @@ class CanvasState {
     return this.isPressed;
   }
 }
+
+class Pan {
```
willgraf, line 175: Did you consider a shared parent `Action` class?

I don't think JavaScript actually has a proper `Interface`, unfortunately, but TypeScript does (for the future).

browser/models.py
```diff
@@ -13,19 +14,21 @@
...
 from helpers import is_npz_file, is_trk_file
 from config import AWS_ACCESS_KEY_ID, AWS_SECRET_ACCESS_KEY
 from imgutils import pngify, add_outlines
 
 
 logger = logging.getLogger('models.Project')  # pylint: disable=C0103
-db = SQLAlchemy()  # pylint: disable=C0103
+# Accessing one-to-many relationships (like project.label_frames) issues a Query, causing a flush
+# autoflush=False prevents the flush, so we still access the db.session.dirty after the query
+db = SQLAlchemy(session_options={"autoflush": False})  # pylint: disable=C0103
```
willgraf, line 31:
```suggestion
db = SQLAlchemy(session_options={'autoflush': False})  # pylint: disable=C0103
```

browser/models.py
```diff
@@ -38,6 +41,31 @@ def compile_pickle_mysql(type_, compiler, **kw):
     return 'LONGBLOB'
 
 
+class MutableNdarray(Mutable, np.ndarray):
+    @classmethod
+    def coerce(cls, key, value):
+        "Convert plain numpy arrays to MutableNdarray."
```
willgraf, line 47:
```suggestion
        """Convert plain numpy arrays to MutableNdarray."""
```

browser/models.py
```diff
@@ -38,6 +41,31 @@ def compile_pickle_mysql(type_, compiler, **kw):
...
+                mutable_array = value.view(MutableNdarray)
+                return mutable_array
+
+            # this call will raise ValueError
+            return Mutable.coerce(key, value)
+        else:
+            return value
+
+    def __setitem__(self, key, value):
+        "Detect array set events and emit change events."
```
willgraf, line 59:
```suggestion
        """Detect array set events and emit change events."""
```

browser/models.py
```diff
@@ -38,6 +41,31 @@ def compile_pickle_mysql(type_, compiler, **kw):
...
+        np.ndarray.__setitem__(self, key, value)
+        self.changed()
+
+    def __delitem__(self, key):
+        "Detect array del events and emit change events."
```
willgraf, line 64:
```suggestion
        """Detect array del events and emit change events."""
```

browser/models.py
```diff
@@ -46,56 +74,92 @@ class Project(db.Model):
...
+        # Load data
         if annotated_key is None:
             annotated_key = get_ann_key(filename)
-
         start = timeit.default_timer()
         trial = self.load(filename, input_bucket, path)
-        current_app.logger.debug('Loaded file %s from S3 in %s s.',
-                                 filename, timeit.default_timer() - start)
+        logger.debug('Loaded file %s from S3 in %ss.',
+                     filename, timeit.default_timer() - start)
```
willgraf, line 116: Why not `current_app`?

browser/blueprints.py
```diff
@@ -173,36 +207,24 @@ def load(filename):
...
 
     # Initate Project entry in database
-    project = Project.create(filename, input_bucket, output_bucket, full_path, rgb)
-    state_start = timeit.default_timer()
-    state = project.state
-    current_app.logger.debug('Got state for "%s" in %s s.',
-                             filename, timeit.default_timer() - state_start)
-
+    project = Project.create(filename, input_bucket, output_bucket, full_path)
+    project.rgb = rgb
```
willgraf, line 211: Why not include `rgb` in the instantiation?

browser/static/js/history.js
```diff
@@ -0,0 +1,135 @@
...
+    }
+    // Redo actions until the end of the group or stack
+    while (true) {
+      action.redo();
+      this.undoStack.push(action);
+      if (this.redoStack.length === 0) break;
+      action = this.redoStack.pop();
+      if (action === 'fencepost') {
+        // Keep fencepost on top of undo stack
+        this.undoStack.push('fencepost');
```
willgraf, line 124:
```suggestion
        this.addFence()
```

browser/static/js/zstack_actions.js
```diff
@@ -0,0 +1,169 @@
...
+ *
+ * @param {string} displayAttr which attribute to change (e.g. frame, feature, or channel)
+ * @param {int} value value to set to attribute
+ */
+function setDisplay(displayAttr, value) {
+  $.ajax({
+    type: 'POST',
+    url: `${document.location.origin}/changedisplay/${project_id}/${displayAttr}/${value}`,
+    success: handlePayload,
+    async: false
```
willgraf, line 140: Have you considered using `async: true` and passing the callback from the Action class?

Synchronous requests can sometimes degrade the user experience.

browser/static/js/zstack_actions.js
```diff
@@ -0,0 +1,169 @@
...
+    });
+  }
+
+  undo() {
+    backendUndo();
+  }
+
+  redo() {
+    backendRedo();
+  }
```
willgraf, line 167: I think these are defined in `main_zstack.js`. I'd feel a little more comfortable if they were in this file instead. Are they required in the other file for some reason?

browser/models.py
```diff
@@ -38,6 +41,31 @@ def compile_pickle_mysql(type_, compiler, **kw):
     return 'LONGBLOB'
 
 
+class MutableNdarray(Mutable, np.ndarray):
```
willgraf, line 44: Can you help me understand what this does/what problem it solves?

willgraf (changes requested):
Great work on this! It looks awesome! I'm still looking through some of the code, but I've left a few comments/suggestions for now. I love the Action interface, though I'm wishing we had some enforcement of it with JavaScript. (Making me think we should look into TypeScript again)

One quick question: Why is there no `track_actions.js`? Is there a way both zstack and track can use the `actions.js` file?

browser/models.py
tddough98: When we change a built-in column type in SQLAlchemy (such as Integer or String), SQLAlchemy knows when the column has been changed and commits the change to the database the next time we call db.session.commit(). 

When using PickleType columns, SQLAlchemy does not have any insight into the mutability/changes to the object it is pickling, and instead only checks the object by reference. When we mutate a numpy array, for instance when replacing a frame with new edited data with `self.frame[..., self.feature] = img`, SQLAlchemy would normally see this object as unchanged as the reference to the array has not changed.

With the Mutable mixin, we can modify the behavior of functions that mutate a numpy array, like`__setitem__`, so that it calls `self.changed()` to notify SQLAlchemy that the data has been changed and needs to be written to the database upon the next commit.

We have not yet used the Mutable mixin with the Labels table, which has two PickleType columns for `cell_ids` and `cell_info`. Therefore, for changes to these columns to be persisted, we copy the dictionaries and reassign them so that the reference has changed and SQLAlchemy is guaranteed to save the changes. There are some TODO comments that show where lack of Mutable for cell_ids/cell_info, and we should aim to use the Mutable mixin for these columns as well so that SQLAlchemy monitors and manages the changes in the background, replacing the code that explicitly copies the data before committing.

browser/models.py
tddough98: Previously when models.py just contained one table that stored just the pickled "state" (ZStackEdit/TrackEdit object), the only log statement in models.py used the logger defined in line 28 in models.py `logger = logging.getLogger('models.Project')`.

As I started moving things from caliban.py into models.py, the current_app logger moved with them, but just before submitting the PR, I noticed the two loggers and went with the original one for consistency.

I don't quite understand the difference between these two loggers and I'd be happy to change all the log statements back to `current_app.logger`, as long as we're consistent in which we're using.

browser/blueprints.py
tddough98: Whether or not a project is in RGB mode does not affect it's creation, so it made sense to simplify the `create` signature and set `rgb` outside of it. The `rgb` setting is more of a cosmetic column that affects what data we display, similar to the current frame/channel/feature.

Also, as we look towards using a file loader interface, we'll likely just pass a file loader instance to `Project.create`, which would pare the signature down to `Project.create(file_loader)`.

browser/static/js/zstack_actions.js
tddough98: They can definitely move into here, I'll make that change now. I think they're separated because I had written them first before making this file.

browser/static/js/canvas.js
tddough98: I've given this an attempt with commit 5e830e5e2e8bc2bb63575f7e8d8f9206943c4d72, let me know if you think this looks better. There's not much logic in the Action class, it just has docstrings and a warning that the function is not implemented.

browser/static/js/zstack_actions.js
tddough98: Not sure if this is exactly the right way to do this, but I changed it to `async: true` with

```
  $.ajax({
    type: 'POST',
    url: `${document.location.origin}/changedisplay/${project_id}/${displayAttr}/${value}`,
    async: true
  }).done(handlePayload);
```

If this looks correct, I'll make some of the other AJAX requests async with the same approach.

tddough98:
> One quick question: Why is there no `track_actions.js`? Is there a way both zstack and track can use the `actions.js` file?

Currently, the tracking app has no undo/redo button. The only changes I made to the tracking app are to use the changed routes on the backend (like `changedisplay` instead of `get_frame`). 

With that said, the actions in `zstack_actions.js` could be used for the tracking app once there is an undo/redo button. Some of the actions like `ChangeChannel` and `ChangeFeature` are not compatible with the tracking app because we assume one channel & one feature for .trk files, but as long as we don't use these actions for tracking, I think it would be fine to rename `zstack_actions.js` to `actions.js`.

browser/static/js/zstack_actions.js
willgraf: Yep, that looks right to me.  Looking at the [docs for ajax](https://api.jquery.com/jquery.ajax/), it looks like we may want to change the `success` calls elsewhere to be `done` as above.

browser/static/js/canvas.js
willgraf: That looks great! You may want to consider making it an abstract class, like in [this SO post](https://stackoverflow.com/a/30560792). It looks like they try to check if a particular method is defined, and throws an error otherwise, which could be helpful here for the 3 required methods. (which is sort of a hacky implementation of Java's abstract classes and interfaces).

browser/static/js/zstack_actions.js
tddough98: Alright, I've changed the requests where `async: true` to call `done` instead of using the `success` parameter.

browser/models.py
willgraf: Ah that makes a lot of sense. I agree, consistency is the most important. Let's keep it as the model/module level logger in that case. Thanks for clearing that up!

browser/models.py
willgraf: Just to confirm, would adding the Mutable mixin to the other columns in the future require another db schema change?

browser/static/js/zstack_actions.js
```diff
@@ -0,0 +1,179 @@
...
+    }).done(handlePayload);
+  }
+
+  undo() {
+    $.ajax({
+      type: 'POST',
+      url: `${document.location.origin}/undo/${project_id}`,
+      async: true
+    });
+  }
```
willgraf, line 169: We don't do anything with the undo response?

tddough98: Oops that's an oversight, fixing now

tddough98:
I've made some significant changes to the undo/redo data models after seeing poor performance on an EB deployment.
Here is an Elastic Beanstalk deployment of Caliban with these latest changes: [http://undo-test.q2qqtvkafv.us-east-2.elasticbeanstalk.com](http://undo-test.q2qqtvkafv.us-east-2.elasticbeanstalk.com)

These changes fall into two categories:
- changing the strategy to represent the project history to store less frames per action
- remodeling the code to better fit [the memento pattern](https://en.wikipedia.org/wiki/Memento_pattern)

# Project History Strategies
First, I want to present a card game metaphor for editing an image stack in Caliban, and then describe three strategies to represent changes in the game.

Imagine a card game with a row of cards laid out on a table, where each "card" corresponds to a frame in a Caliban image stack. Each turn, the player lays down a new card one or more of the card stacks, just like how with each action in Caliban we can edit one or more frames. To distinguish each action, we place a sheet of clear plastic on the table before placing the cards in an action. Over time, we build a stack of plastic & cards with the cards placed during a particular action isolated in their own layer. At any point in time, we can peer through the layers of plastic and cards to see the last card placed at any position, showing us the current game state.

When we want to undo the most recent action, we remove the top sheet of plastic along with any cards on it, and to redo the action, we put the sheet & cards back into place. In this metaphor, when we remove a card, we can peer through the layers of plastic to see what the last card under it was, in effect immediately updating the game state to the correct past state. However, updating this game state means that we need to look through the past action layers until the last version is visible.

The key challenge revealed by this metaphor is that when undoing/redoing actions, we need to have a strategy to know where the correct past/future version of a "card" or frame is stored.

## Strategies

Here, I'll describe three strategies to meet this challenge, highlight some pros & cons, and explain which strategy we use.

1. Record the unedited & edited versions of the frames an action edits
2. Record the initial frames & the edited version of the frames an action edits
3. Record the final frames & the unedited version of the frames an action edits

1. First, whenever an action edits a frame, we can record the unedited & edited copies of the frame. This option makes accessing the correct versions trivial when undoing as they are always stored directly in the frame, but it requires two copies of the each changed frame, which slows down each action with more database writes.

3. Second, we can store an initial copy of the project and record the edited versions of edited frames.
This strategy most closely aligns with the card game metaphor above, as the cards first laid on the table representing the initial state and the layers with the placed cards 
This strategy allows us to store only one frame copy per frame change, and makes redoing actions easy, as the changed frames are always stored directly with the action. On the other hand, undo becomes more challenging as if we undo an action where the last version of a changed frame is stored many actions back, we need to proactively seek out that version. We've now accounted for this search, which is implemented in the `before_frames` property of an Action. There is a placeholder test `test_undo_action_with_frame_not_in_previous_action` to check for this issue, but it still needs to be written. 

2. Third, we can record the final state of the project and the unedited copies of each edited frame. This setup is similar to the metaphor above, except we keep the current game state on top and keep the old cards hidden underneath. Previously, this PR used this strategy to store the project history. While this strategy can correctly represents all the changes, it has the worst performance, as every action changes the final project state and we have to rewrite all the frames in an action everytime we update the project. Additionally, the PR did not account for looking for changed frames many actions ahead, causing a bug with redo where if the next version of a frame is stored many actions ahead, that version will not be loaded until we reach that action. There is also a placeholder test `test_redo_action_with_frame_not_in_next_action` to check for this issue, but it still needs to be written. 

We now opt for strategy 2, as we can take some extra time to make a copy of the frame when loading the file, and then only write one copy of changed frames in each action.

# [Memento pattern](https://en.wikipedia.org/wiki/Memento_pattern)

Lastly, we've made some changes to better follow the memento pattern. We don't follow the memento pattern exactly as we've combined the caretaker and the originator, so a Project manages its own undoing & redoing. Action serve as the memento and record the internal state needed, edited frames and labels, to restore the project to a past point in time.

willgraf (changes requested):
This is great! I think the backend functionality is good to go, but I have a couple of nitpicky whitespace and CSS changes.

browser/static/js/main_zstack.js
```diff
@@ -1036,68 +960,86 @@ function handleMousemove(evt) {
...
+
+function displayUndoRedo() {
+  let canvasElement = document.getElementById('canvas');
+  let undoButton = document.getElementById('undo');
+  undoButton.hidden = false;
+  undoButton.style.width = canvasElement.width / 2 + 'px';
+  
+  let redoButton = document.getElementById('redo');
+  redoButton.hidden = false;
+  redoButton.style.width = canvasElement.width / 2 + 'px';
```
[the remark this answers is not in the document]
tddough98: I put both of these (setting hidden to false & setting the width) on the javascript side because they are dynamic behaviors. We hide the undo/redo buttons until the image is loaded, and we adjust the width of the buttons ti match the canvas width whenever we resize the window as the canvas resized by
```
    // resize the canvas every time the window is resized
    window.addEventListener('resize', function() {
      waitForFinalEvent(() => {
        mode.clear();
        setCanvasDimensions(payload.dimensions);
        brush.refreshView();
        displayUndoRedo(); // match the button width to the canvas width
      }, 500, 'canvasResize');
    });
```

Is there a way to dynamically adjust these attributes just with CSS/Materialize classes?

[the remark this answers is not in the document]
willgraf: Yes, materialize can dynamically size elements according to their column sizes. However, I notice now that we are changing the size of the canvas with JS and linking the buttons to that canvas size. This is not nearly as easy to resolve with CSS, so let's not worry about it.

browser/models.py
tddough98: I don't think it would require a schema change as the Mutable mixin is just on the ORM side and wouldn't change the column declarations on the SQL side. Still, if it does require a schema change, I'd imagine it would be a straightforward migration as we wouldn't be adding or moving columns, just changing how these columns behave.

geneva-miller:
I'm interacting with the instance and have a few things I've noticed:

- drawing is a little buggy, clicking once with the brush should be enough to draw a spot but it looks like it needs a double click to send the action

- I'm not a fan of the async. For someone moving reasonably quickly through frames, the behavior seems a little out of whack especially in pixel-editing mode. I think the main thing is that with async, you can still (kind of) interact with the out-of-date frame while you're waiting for an action to finish and update the page. While this does make things feel a little smoother, I don't think users should be able to interact with the file if their view is not current. If we could disable further interactions while waiting I would be okay with async I think

- I'm quite liking the undo functionality! I am quite happy with the location, size, and resizing of the undo/redo buttons

browser/models.py
```diff
@@ -156,166 +222,248 @@ def create(filename, input_bucket, output_bucket, path, rgb=False):
...
+        Records the effects of the action in the Action table.
         """
         start = timeit.default_timer()
+        if self.action.labels_changed:
+            self.labels.update()
         db.session.commit()
         logger.debug('Updated project %s in %ss.',
                      self.id, timeit.default_timer() - start)
 
     def finish(self):
```
geneva-miller, line 251: do rows in the action table need to get cleared upon project finish?

browser/static/js/actions.js
```diff
@@ -0,0 +1,179 @@
...
+}
+
+/** Action to change the viewed frame. */
+class ChangeFrame extends Action {
+  constructor(mode, frame) {
+    super();
+    this.mode = mode;
+    this.oldValue = current_frame;
+    // frame mod max_frames
+    this.newValue = ((frame % max_frames) + max_frames) % max_frames;
```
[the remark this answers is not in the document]
tddough98: `%` is [remainder](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Remainder) in Javascript, not mod and it can be negative, e.g. `-1 % 10 == -1` rather than 9. This expression implements mod in terms of remainder, taken from this [StackOverflow answer](https://stackoverflow.com/questions/4467539/javascript-modulo-gives-a-negative-result-for-negative-numbers).

browser/static/js/zstack_actions.js
```diff
@@ -0,0 +1,177 @@
+class ToggleEdit extends Action {
```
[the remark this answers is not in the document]
tddough98: You're right, good catch

browser/models.py
tddough98: We don't clear them out currently, as Erick & Dave mentioned interest in building an "action dataset" for reinforcement learning/other uses. As we only stored the changed frames in the action history, this should be relatively lightweight, but we should monitor the database size to see if this become unwieldy.

browser/static/js/main_zstack.js
```diff
@@ -51,144 +21,139 @@ class Mode {
...
       // toggle rendering_raw
       rendering_raw = !rendering_raw;
       render_image_display();
-    } else if (key === '0') {
+    } else if (evt.key === '0') {
       // reset brightness adjustments
-      adjuster.resetBrightnessContrast();
-    } else if ((key === 'l' || key === 'L') && rgb && !edit_mode) {
+      let resetBrightnessContrast = new ResetBrightnessContrast(adjuster);
+      actions.addAction(resetBrightnessContrast);
```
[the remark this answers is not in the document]
tddough98: That makes senses, I'll make that change

browser/static/js/main_zstack.js
```diff
@@ -1189,17 +1132,18 @@ function startCaliban(filename, settings) {
         mode.clear();
         setCanvasDimensions(payload.dimensions);
         brush.refreshView();
+        displayUndoRedo();
       }, 500, 'canvasResize');
     });
 
     window.addEventListener('keydown', (evt) => {
-      mode.handle_key(evt.key);
+      mode.handle_key(evt);
```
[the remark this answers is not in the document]
tddough98: I needed to make this changed for Ctrl+Z support. Whether the Ctrl key is pressed is stored in `evt.ctrlKey`, so we pass only `evt.key` to the handler, we can't tell if Ctrl was also pressed.

browser/static/js/main_zstack.js
```diff
@@ -51,144 +21,139 @@ class Mode {
...
   }
 
   // these keybinds apply regardless of
   // edit_mode, mode.action, or mode.kind
-  handle_universal_keybind(key) {
-    if (!rgb && (key === 'a' || key === 'ArrowLeft')) {
+  handle_universal_keybind(evt) {
+    if ((evt.ctrlKey || evt.metaKey) && evt.shiftKey && (evt.key === 'Z' || evt.key === 'z')) {
+      redo();
+    } else if ((evt.ctrlKey || evt.metaKey) && evt.key === 'z') {
```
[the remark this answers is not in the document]
tddough98: Hmmmm I buy that. I added the 'Z' or 'z' condition because just 'Z' wasn't working, but the Caps Lock issue is a good point. At some point, we should move to use [Mousetrap](https://craig.is/killing/mice)to handle the keybinds, which manages the case, Ctrl vs. Command for Windows/Mac, etc. For now I'll add the 'Z' or 'z' check to the second condition too.

browser/static/js/main_zstack.js
```diff
@@ -51,144 +21,139 @@ class Mode {
...
+      undo();
+    } else if (!rgb && (evt.key === 'a' || evt.key === 'ArrowLeft')) {
```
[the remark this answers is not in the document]
tddough98: Yes, the backend can handle multiframe RGB files, I'll make this change.

There are a couple of other places where we check `!rgb`, namely:
- `if (!rgb && evt.key === 'h')` for toggling highlight
- `if (!rgb && evt.key === 'i')` for toggling invert
- `if (!rgb && settings.pixel_only && (evt.key === 'l' || evt.key === 'L'))` for toggling display labels
- `if (evt.key === 't' && !rgb)` for thresholding
- `if (evt.key === 'p' && !rgb)` for predicting zstack labels
- `if (evt.key === 'f' && !rgb)` for filling holes
- `if (!rgb && evt.key === 'c')` for creating a new cell stack
- `if (!rgb && evt.key === 's')` for swapping labels
- `if (evt.key === 'w' && !rgb)` for watershed

I'm not sure what the issues would be for enabling these features for RGB mode, but I want to draw attention to them if there's any others that should be enabled.

browser/static/js/actions.js
```diff
@@ -0,0 +1,179 @@
...
+    this.do()
+  }
+
+  redo() {
+    this.do()
+  }
+}
+
+/** Action to change the viewed frame. */
+class ChangeFrame extends Action {
```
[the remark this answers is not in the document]
tddough98: I don't think that this should be logic inside the ChangeFrame class, because an Action can't tell the history if it was actually performed or not. Instead, we shouldn't create & perform the action in the first place. I'll add conditions to the key handlers that check that `max_frames > 1`, etc.

browser/static/js/main_zstack.js
```diff
@@ -200,66 +165,57 @@ class Mode {
...
-      // cycle highlight to next label
-      this.highlighted_cell_one = this.increment_value(
-        this.highlighted_cell_one,
-        1,
-        maxLabelsMap.get(this.feature)
-      );
+    } else if (evt.key === ']' && this.highlighted_cell_one !== -1) {
+      // cycle highlight to next label (skipping 0)
+      let maxLabel = 1 + maxLabelsMap.get(this.feature);
+      this.highlighted_cell_one = (this.highlighted_cell_one + 1) % (maxLabel) + 1;
```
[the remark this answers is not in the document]
tddough98: Ah I see, I was messing around with this in edit mode & I see this causes issues in whole label mode. Looks good now

browser/static/js/main_zstack.js
```diff
@@ -200,66 +165,57 @@ class Mode {
...
+    } else if (evt.key === 'p' && !rgb) {
       // iou cell identity prediction
       this.kind = Modes.question;
       this.action = 'predict';
       this.prompt = 'Predict cell ids for zstack? / S=PREDICT THIS FRAME / SPACE=PREDICT ALL FRAMES / ESC=CANCEL PREDICTION';
       render_info_display();
-    } else if (key === '[' && this.highlighted_cell_one !== -1) {
+    } else if (evt.key === '[' && this.highlighted_cell_one !== -1) {
       // cycle highlight to prev label
       this.highlighted_cell_one = this.decrement_value(
```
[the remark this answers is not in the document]
tddough98: Changing this to look like the increment code would be
```let numLabels = maxLabelsMap.get(this.feature) - 1;
      this.highlighted_cell_one = (((this.highlighted_cell_one - 2) % numLabels) + numLabels) % numLabels + 1;```
which is a little messy. 

With increment, we can easily create the cycling just with `highlighted % maxLabels + 1`, but with decrementing, we run up against the `%` remainder returning negative values & not wanting to highlight 0 as a label. I think it might be better to keep `decrement_value` around for this.

File: browser/static/js/main_zstack.js [hunk 26]
```diff
@@ -751,14 +697,14 @@ function upload_file(cb) {
...
     async: true
-  });
+  }).done(cb);
 }
 
 function changeZoom(dzoom) {
-  state.changeZoom(dzoom, state.canvasPosX, state.canvasPosY);
-  updateMousePos(state.rawX, state.rawY);
+  zoom = new Zoom(canvas, dzoom);
+  updateMousePos(canvas.rawX, canvas.rawY);
```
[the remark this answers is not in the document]
tddough98: Hmmm I'm struggling with this one... updateMousePos interacts with multiple global objects (both canvas and brush), and is used both by moving the mouse and by Zooming.

Given that undo/redoing clears the mode (& the selected labels), I'm not sure if it's possible to do an action immediately after undoing without moving the mouse and updating the mouse position. The brush also shows the old location until its updated, so it is still clear which label is selected, even if the cursor is not lined up with it.

I'm going to not address this issue because it would require too much restructuring of the global objects and their interactions.

browser/static/js/main_track.js
```diff
@@ -715,7 +714,8 @@ function load_file(file) {
     url: `${document.location.origin}/load/${file}`,
     success: function (payload) {
       max_frames = payload.max_frames;
-      scale = payload.screen_scale;
+      //scale = payload.screen_scale;
+      scale = 2;
```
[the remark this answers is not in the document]
tddough98: Let's use the scale_factor from the payload, I think this was changed as a stopgap when the scale_factor from the payload wasn't working correctly.

browser/static/js/main_zstack.js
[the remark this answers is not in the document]
tddough98: I took another look, and I think there may be a reasonable solution by updating the mouse position after any undo/redo, not just Zooms. I got tripped up trying to record the mouse history in the Zoom, and I think this is something that should not be tracked by undo/redo, just made consistent again. I think with this change the behavior works as expected.

browser/static/js/main_zstack.js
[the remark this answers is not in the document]
geneva-miller: - Highlighting: let's turn on the option to toggle the highlight, but also start both rgb and non-rgb modes with highlighting on.
- Inverting: should not be an option for rgb
- "L" for toggling display labels is pretty niche and only gets used if "pixel-only" editing is specified--let's leave this for now but this is something that would get addressed if we decoupled image display from editing mode
- Thresholding: keep as non-rgb only for now. we could allow it in the future for rgb images probably if we specified which channel to threshold on, perhaps when adding in individual channel brightness adjustment
- Predicting labels: only when rgb is multiframe-friendly
- Filling holes: not sure why this is set to rgb only, it should be available in rgb mode as well, we just need to make sure it doesn't ever get called at the same time as change_feature
- Creating new cell stack: when rgb is multiframe-friendly
- Swapping labels: we could make this an rgb feature, it just didn't have any utility at the time for Noah's work
- Watershed: same as thresholding

browser/static/js/main_zstack.js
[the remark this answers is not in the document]
geneva-miller: Oh awesome. I hadn't realized that ctrl wasn't stored in `evt.key` although I think that is the reason why zooming uses alt instead of ctrl. I would love to make more use of the ctrl key in future work for intuitive keybinds.

browser/static/js/main_zstack.js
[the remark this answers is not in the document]
geneva-miller: I'm not sure what is misbehaving here, but the behavior is a little odd. When adjusting the brightness, then resetting, then undoing, there are two undo steps as expected of the fenced action. But the first undo (undoing reset) doesn't change the image display as expected. Redoing both adjustments does work as expected though.

browser/static/js/main_zstack.js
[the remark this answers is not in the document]
geneva-miller: You're right, the implementation is messier than I expected (I didn't look up `%`'s behavior in js, my bad). With that in mind I am fine keeping `decrement_value` in for now. If it had been a simple and clean change I would have liked to change it at the same time as `increment_value` but as it is, let's just consider revisiting it at some point.

browser/static/js/adjust.js
```diff
@@ -384,3 +343,97 @@ class ImageAdjuster {
...
+    this.adjuster = adjuster;
+    this.brightness = adjuster.brightness;
+    this.contrast = adjuster.contrast;
+  }
+
+  do() {
+    adjuster.brightness = 0;
+    adjuster.contrast = 0;
+    adjuster.rawLoaded = false;
+    adjuster.contrastRaw();
```
geneva-miller, line 412:
```suggestion
    this.adjuster.brightness = 0;
    this.adjuster.contrast = 0;
    this.adjuster.rawLoaded = false;
    this.adjuster.contrastRaw();
```

browser/static/js/adjust.js
```diff
@@ -384,3 +343,97 @@ class ImageAdjuster {
...
+  }
+
+  undo() {
+    adjuster.brightness = this.brightness;
+    adjuster.contrast = this.brightness;
+    adjuster.rawLoaded = false;
+    adjuster.contrastRaw();
```
geneva-miller, line 419:
```suggestion
    this.adjuster.brightness = this.brightness;
    this.adjuster.contrast = this.contrast;
    this.adjuster.rawLoaded = false;
    this.adjuster.contrastRaw();
```
fixes the unexpected undo behavior I was seeing in user testing

geneva-miller:
The other thing I'm looking into is making sure that rgb mode is multi-frame friendly (using [this file](http://undo-test.q2qqtvkafv.us-east-2.elasticbeanstalk.com/caliban-input__caliban-output__janelia_demo__HeLa-S3_janelia_demo_version.npz?rgb=True)). Once I change frames in the file, though, I get single channel pngs returned to me from the server and I'm not sure why.

browser/static/js/main_zstack.js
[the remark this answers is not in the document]
tddough98: Looking back at [this resource on mod vs remainder](https://web.archive.org/web/20090717035140if_/javascript.about.com/od/problemsolving/a/modulobug.htm), I tried adding the mod function to the Number prototype, so mod becomes `x.mod(divisor)` and the code to decrement while skipping 0 becomes
```
let numLabels = maxLabelsMap.get(this.feature) - 1;
this.highlighted_cell_one = (this.highlighted_cell_one + numLabels - 2).mod(numLabels) + 1;
```
which is still a bit clumsy, but now more clearly using mod to cycle values. These lines plus a quick comment `// cycle down, skipping 0` I think is better than the decrement value function, which had three argument that a reader would have to cross reference. Adding mod to the Number prototype also cleans up some of the actions like ChangeFrame to have lines like `frame.mod(numFrames)` instead. Also, I'm changing the variables used as a divisor to be more consistent, in particular `max_frames` --> `numFrames, `channelMax` --> `numChannels`, and `feature_max` --> `numFeatures`. I'll make a separate comment about this change so it's more prominent, but here's a heads up in case you see a ton of changed lines in the javascript files with these variables.

I think better solution down the road could rework `highlighted_cell_one` to implicitly offset by 1 (so a value of 0 would highlight the first label), making the decrementing/incrementing simplify to `(highlighted_cell_one - 1).mod(numLabels)` and `(highlighted_cell_one + 1).mod(numLabels)`. This would be a more sweeping change, so I'm holding off for now.

geneva-miller:
@tddough98 The current instance gets stuck on loading files (displays "file loading" page but then returns 500 error). Could you relaunch this branch to undo-test? I would like to double check the prediction relabeling and have a working instance up while I review the code.

geneva-miller (commented):
One small thing I noticed. Undo functionality and related changes all look and work well, so I'm otherwise ready to approve this PR.

browser/static/js/actions.js
```diff
@@ -0,0 +1,176 @@
...
+    this.mode = mode;
+    this.oldValue = current_frame;
+    this.newValue = frame.mod(numFrames);
+  }
+
+  do() {
+    setDisplay('frame', this.newValue);
+    current_frame = this.newValue;
+    if (this.mode.action !== '') { this.mode.clear() };
+    render_info_display();
```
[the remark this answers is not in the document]
tddough98: Ah I see the callback in `setDisplay` is `handlePayload`, which calls `render_image_display`, which calls `render_info_display`. The only issue I could see is that we're changing some info after the setDisplay call, like the mode and current_frame, which would be a bug if setDisplay made a synchronous call, but probably not for async. I'll move these info-changing statements above `setDisplay` so it's never an issue and remove the redundant `render_info_display` call.

geneva-miller (approved):

browser/static/js/actions.js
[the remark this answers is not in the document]
geneva-miller: Tested with `test.npz` which has multiple frames, features, and channels. Looks good!

browser/templates/base.html
```diff
@@ -66,9 +66,15 @@
...
     <!-- Load Custom JavaScript -->
+    <script type="text/javascript" src="{{ url_for('static', filename='js/history.js') }}"></script>
     <script type="text/javascript" src="{{ url_for('static', filename='js/adjust.js') }}"></script>
     <script type="text/javascript" src="{{ url_for('static', filename='js/brush.js') }}"></script>
     <script type="text/javascript" src="{{ url_for('static', filename='js/canvas.js') }}"></script>
+    <script>
+      Number.prototype.mod = function(n) {
+        return ((this%n)+n)%n;
+      };
+    </script>
```
willgraf, line 77: What does this do?

willgraf (approved):
💯

PR #213, opened 2020-11-06
Use parallel Coveralls reports in GitHub actions

willgraf:
## What
* Use parallel jobs for Coveralls

## Why
* Provide better insight into test coverages across different versions.


.github/workflows/browser-tests.yaml
```diff
@@ -22,15 +22,39 @@ jobs:
...
+    - name: Cache pip
+      uses: actions/cache@v2
+      with:
+        # This path is specific to Ubuntu
+        path: ~/.cache/pip
+        # Look to see if there is a cache hit for the corresponding requirements file
+        key: ${{ runner.os }}-pip-${{ hashFiles('requirements.txt') }}
+        restore-keys: |
+          ${{ runner.os }}-pip-
+          ${{ runner.os }}-
```
tddough98, line 34: Why are there two keys here? I don't see any issue with it, just curious. I saw this [question](https://github.community/t/pip-dependencies-not-getting-cached-even-after-successful-restore/126672) using the same keys in an answer, but it doesn't explain why this works.

tddough98 (approved):
Looks good. Both caching pip and checking coverage across versions seem useful.

.github/workflows/browser-tests.yaml
willgraf: I don't have a great answer for you; it was featured that way in the examples.

PR #217, opened 2020-12-09
File loader schema bridge

tddough98:
## What
This PR adds the schema changes needed by the file-loader PR #214  to the master branch without functionality of #214. The schema changes all affect the Projects table and include:

- using only a `path` column instead of both a `filename` and a `path` column
- remove the `output_bucket` column
- add a `token` column that gives Projects a unique 12 character base64 identifier
- add a `source` column that records where a project was loaded from (currently always `s3`)

In line with these schema changes, we also change:

- the internal `load` and `upload_file` routes now include a `bucket` component like `/load/<bucket>/<projectID>`
- the `create` function and Project constructor now only have two arguments, `path` and `bucket`, instead of `filename`, `path`, `input_bucket`, and `output_bucket`.
- the tool and shortcut routes now share a `make_settings` helper function to prepare the settings dictionary sent to `tool.html`
- the settings dictionary now includes `input_bucket`, `output_bucket`, `filename`, `filetype`, and `title`. The first two are new, and the last three previously were passed alone to `tool.html`

Here is a EB deployment to test these changes: [http://bridge-test.q2qqtvkafv.us-east-2.elasticbeanstalk.com/](http://bridge-test.q2qqtvkafv.us-east-2.elasticbeanstalk.com/). This deployment is sharing a database with the current file-loader deployment to confirm that their schemas are compatible.

## Why
Both  #212 and #214 introduces changes to the database schema. #212 has been merged in and provides performance and usability improvements, so we want to deploy these features as soon as possible. However, #214 is still under development and review. If we were to deploy #212 with no changes, we would have the migrate the database to deploy #214, but with these additional changes, we have the minimum support needed to merge and deploy #214 with no migration.

### DIFF
--- a/browser/blueprints.py
+++ b/browser/blueprints.py
@@ -213,7 +213,6 @@ def load(bucket, filename):
     # arg is 'false' which gets parsed to True if casting to bool
     rgb = request.args.get('rgb', default='false', type=str)
     rgb = bool(distutils.util.strtobool(rgb))
-    bucket = request.args.get('input-bucket', default=S3_INPUT_BUCKET, type=str)
 
     if not is_trk_file(path) and not is_npz_file(path):
         ext = os.path.splitext(path)[-1]
--- a/browser/blueprints.py
+++ b/browser/blueprints.py
@@ -48,11 +48,13 @@ class InvalidExtension(Exception):
         rv['message'] = self.message
         return rv
 
+
 @bp.errorhandler(InvalidExtension)
 def handle_invalid_usage(error):
     response = jsonify(error.to_dict())
     response.status_code = error.status_code
     return response
+
 
 @bp.errorhandler(Exception)
 def handle_exception(error):
@@ -291,12 +293,12 @@ def get_edit(project):
 def make_settings(filename):
     """Returns a dictionary of settings to send to the front-end."""
     folders = re.split('__', filename)
-    
+
     input_bucket = folders[0] if len(folders) > 1 else S3_INPUT_BUCKET
     output_bucket = folders[1] if len(folders) > 2 else S3_OUTPUT_BUCKET
     start_of_path = min(len(folders) - 1, 2)
     path = '__'.join(folders[start_of_path:])
-    
+
     rgb = request.args.get('rgb', default='false', type=str)
     pixel_only = request.args.get('pixel_only', default='false', type=str)
     label_only = request.args.get('label_only', default='false', type=str)
--- a/browser/blueprints.py
+++ b/browser/blueprints.py
@@ -22,7 +22,7 @@ from werkzeug.exceptions import HTTPException
 from helpers import is_trk_file, is_npz_file
 from models import Project
 from caliban import TrackEdit, ZStackEdit, BaseEdit, ChangeDisplay
-
+from config import S3_INPUT_BUCKET, S3_OUTPUT_BUCKET
 
 bp = Blueprint('caliban', __name__)  # pylint: disable=C0103
 
@@ -55,13 +55,15 @@ def upload_file(project_id):
     if not project:
         return jsonify({'error': 'project_id not found'}), 404
 
+    bucket = request.args.get('output-bucket', default=S3_OUTPUT_BUCKET, type=str)
+
     # Call function in caliban.py to save data file and send to S3 bucket
     edit = get_edit(project)
     filename = project.filename
     if is_trk_file(filename):
-        edit.action_save_track()
+        edit.action_save_track(bucket)
     elif is_npz_file(filename):
-        edit.action_save_zstack()
+        edit.action_save_zstack(bucket)
 
     # add "finished" timestamp and null out PickleType columns
     project.finish()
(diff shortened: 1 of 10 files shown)


browser/blueprints.py
```diff
@@ -33,6 +33,29 @@ def health():
...
+@bp.errorhandler(InvalidExtension)
+def handle_invalid_usage(error):
+    response = jsonify(error.to_dict())
+    response.status_code = error.status_code
+    return response
+
+
 @bp.errorhandler(Exception)
 def handle_exception(error):
     """Handle all uncaught exceptions"""
```
willgraf, line 61: How does Flask cascade these errors? Is it determined by the declaration order?

browser/blueprints.py
```diff
@@ -294,46 +273,58 @@ def shortcut(filename):
...
         return ZStackEdit(project)
     elif is_trk_file(filename):
         # don't use RGB mode with track files
         return TrackEdit(project)
     return BaseEdit(project)
+
+
+def make_settings(filename):
+    """Returns a dictionary of settings to send to the front-end."""
+    folders = re.split('__', filename)
```
willgraf, line 295: Is this different than `filename.split('__')`?

browser/models.py
```diff
@@ -212,22 +217,26 @@ def get(project_id):
...
         """
         start = timeit.default_timer()
-        new_project = Project(filename, input_bucket, output_bucket, path)
+        new_project = Project(path, bucket)
+        # Assign a unique 12 character base64 token to the project
+        while True:
+            token = token_urlsafe(9)  # 9 bytes is 12 base64 characters
+            if not db.session.query(Project).filter_by(token=token).first():
+                new_project.token = token
+                break
```
willgraf, line 239: Just keep trying to make a random token until we find one that is unused?

Also, is there a specific reason you chose 12 chars long?

geneva-miller:
I tried loading a file from the output bucket by changing the url, but the file from the input bucket was loaded.

browser/static/js/main_track.js
```diff
@@ -838,7 +840,7 @@ function prepare_canvas() {
 function action(action, info, frame = current_frame) {
   $.ajax({
     type:'POST',
-    url:`${document.location.origin}/action/${project_id}/${action}/${frame}`,
+    url:`${document.location.origin}/edit/${project_id}/${action}/${frame}`,
```
willgraf, line 843: Was this a missed bugfix?

willgraf (commented):
This all looks good to me. Mostly just changing the constructor calls to only have `path`. I left a few questions just for clarification, and I will test out the deployment now.

browser/static/js/main_track.js
tddough98: Yes, this was missed from the undo PR. The action route was renamed to edit

browser/models.py
tddough98: Yes, keep making random ones until it's unused. This should happen immediately, as the probability of a collision when we have ~60,000 projects (which I think is how are in the current db) and use 12 character IDs is `2^16 / (2^8)^12` = 2^(-80)`.

The number of base 64 characters needed to represent `n` bytes is `n * 4 / 3`, so I wanted to choose a multiple of 3, and youtube uses 11 characters, so I chose 12.

browser/blueprints.py
tddough98: I don't think so. Change it to reduce the number of imports?

tddough98:
> I tried loading a file from the output bucket by changing the url, but the file from the input bucket was loaded. The `inputBucket` variable on the frontend seems to be set correctly to `caliban-output`

Should be fixed now, the bucket argument in the `load` route was being overwritten by a URL parameter default so it always became the default input, but now that line is gone.

browser/blueprints.py
tddough98: From [https://flask.palletsprojects.com/en/1.1.x/errorhandling/#handling](https://flask.palletsprojects.com/en/1.1.x/errorhandling/#handling):

>When an exception is caught by Flask while handling a request, it is first looked up by code. If no handler is registered for the code, it is looked up by its class hierarchy; the most specific handler is chosen.

browser/blueprints.py
tddough98: actually `re` is also used in line 211 
```
path = re.sub('__', '/', filename)
```
but we could use `filename.replace('__', '/')` instead to get rid of all `re` calls

browser/blueprints.py
```diff
@@ -294,46 +272,58 @@ def shortcut(filename):
...
+
+    input_bucket = folders[0] if len(folders) > 1 else S3_INPUT_BUCKET
```
geneva-miller, line 296: Most files are stored in subfolders, so I think this logic for the bucket names could create confusion. Maybe we could use `S3_INPUT_BUCKET` if `folders[0] == ''`? The urls to create those paths would not look pretty but it would avoid bucket/subfolder ambiguity

browser/blueprints.py
willgraf: Doesn't matter to me, `re` is the standard library. Just wanted to make sure as I think the previous version used `str.split()`.

geneva-miller:
Tested file loading and submission with available test files, mostly focused on urls that include both input and output bucket info. Loading and submitting files works as expected when bucket info is included in url, as well as when no bucket info or subfolders are included in the url.

(Sequence of tests: Load [this file](http://bridge-test.q2qqtvkafv.us-east-2.elasticbeanstalk.com/test.npz), modify, and submit to confirm that bucket defaults work. Output file can be loaded from [here](http://bridge-test.q2qqtvkafv.us-east-2.elasticbeanstalk.com/caliban-output__caliban-output__test.npz) to confirm that your changes were submitted. Also tested with [this copy](http://bridge-test.q2qqtvkafv.us-east-2.elasticbeanstalk.com/caliban-output__caliban-output__test__test.npz) of that file, which is located within a subfolder. This file is loaded from and saved to the output bucket. Tested by opening file, making change, submitting, and reloading page to see the changes I had submitted. Finally, loaded some actual files from the Anolytics job currently running to double check that their file loading will work as expected, but those aren't test files so I'm not going to link them.)

The annotation jobs we have running have the input and output buckets specified in the urls, and those continue to work as expected. So, I'm not too concerned about url edge cases like trying to pass in subfolders without specifying bucket names in this PR.

browser/blueprints.py
willgraf: I agree that this is a little unclear, but as mentioned below the URLs/sub folders will probably not be an issue until the file loader is merged in. Maybe we could just add a TODO here to take care of it in the next PR for file uploading?

willgraf (approved):
I'm tentatively approving since it sounds like @geneva-miller has successfully run through some test workflows and everything works as expected.

There is more work to do with regards to URLs, but since that work is in addition to the current usage of the URLs, I think this is OK to merge in to master for immediate use by annotators.

browser/blueprints.py
geneva-miller: Yep, fine by me

PR #218, opened 2020-12-23
Erode dilate contour master

tddough98:
## What
This PR is a hotfix for three additional actions in DeepCell Label: erode, dilate, and active contour. The names of these actions on the frontend are shrink, expand, and autofit, respectively. The keybinds are q (for erode), Q (for dilate), and m (for autofit), then a Space to confirm the action. They are only available in whole-label mode and require a single label to be selected.

A test instance of this PR is available here: [new-actions-test.q2qqtvkafv.us-east-2.elasticbeanstalk.com](https://new-actions-test.q2qqtvkafv.us-east-2.elasticbeanstalk.com)

## Why
These actions will allow for cleaner, faster adjustment of labels without a brush.

This PR does not include tests for the new actions, but the tests are written in branch erode-dilate-active-contour (see commit dfde2d228a395db6813515075e097cf9a805dd21), as these tests use the new DummyLoader features from the file-loader branch. You can see on the linked commit that the tests are passing for these actions. We're aiming to merge in these actions without the rest of the file-loader branch so that we can start using them with Anolytics without any hiccups the file loaders might cause. I'll add these tests to the file-loader branch as soon as this is merged in.


browser/caliban.py
```diff
@@ -473,6 +475,111 @@ def action_threshold(self, y1, x1, y2, x2, label):
...
+        pixel_count = np.count_nonzero(safe_overlay == label)
+        if pixel_count < 20:
+            safe_overlay = np.copy(self.frame[y1:y2, x1:x2, self.feature])
+
+        # put it back in the full image so can use centroid coords for post-contour cleanup
+        full_frame = np.copy(self.frame[..., self.feature])
+        full_frame[y1:y2, x1:x2] = safe_overlay
+
+        # avoid automated label cleanup if the centroid (flood seed point) is of background
+        # TODO: perhaps this should be "if (seed point value) != label" instead
```
geneva-miller, line 529: Let's make this change, redeploy, and do a quick sanity test of functionality. This would only make a difference in some very weird edge cases but it ought to be quick and easy to change now.

geneva-miller (commented):
I've tested this on one of the files that Anolytics currently has to work on (a few other files as well) and it works just as expected. Hoping to get this in ASAP so I can get annotators to start using this.

browser/caliban.py
```diff
@@ -473,6 +475,110 @@ def action_threshold(self, y1, x1, y2, x2, label):
...
+
+        # if label is adjacent to another label, don't let that interfere
+        img_erode = np.where(img_ann == label, label, 0)
+        # erode the label
+        img_erode = erosion(img_erode, square(3))
+
+        # put the label back in
+        img_ann = np.where(img_ann == label, img_erode, img_ann)
+
+        in_modified = np.any(np.isin(img_ann, label))
```
willgraf, line 560: If `label` is just an integer I think just checking equality is faster than `np.isin`.

```suggestion
        in_modified = np.any(img_ann == label)
```

browser/caliban.py
```diff
@@ -473,6 +475,110 @@ def action_threshold(self, y1, x1, y2, x2, label):
...
+        # contoured area should get original label value
+        contoured_label = np.where(contoured, label, 0)
+        # contours tend to fit very tightly, a small expansion here works well
+        contoured_label = dilation(contoured_label, disk(3))
+
+        # don't want to leave the original (un-contoured) label in the image
+        label_img = np.where(label_img == label, 0, label_img)
+
+        # never overwrite other labels with new contoured label
+        safe_overlay = np.where(label_img == 0, contoured_label, label_img)
```
willgraf, line 516: Are these the same commands? I think the single `np.where` is faster.

```suggestion
        # don't want to leave the original (un-contoured) label in the image
        # never overwrite other labels with new contoured label
        cond = np.logical_or(label_img == label, label_img == 0)
        safe_overlay = np.where(cond, contoured_label, label_img)
```

willgraf (changes requested):
A few numpy-related performance suggestions.

browser/caliban.py
geneva-miller: I investigated this with timeit and didn't find a difference. So I think it's down to the style you prefer here

browser/caliban.py
```diff
@@ -473,6 +475,110 @@ def action_threshold(self, y1, x1, y2, x2, label):
...
+        # normalize input 2D frame data values to range [0.0, 1.0]
+        adjusted_raw_frame = Normalize()(self.raw_frame[..., self.channel])
+        predict_area = adjusted_raw_frame[y1:y2, x1:x2]
+        # import pdb; pdb.set_trace()
+
+        # returns 1 where label is predicted to be based on contouring, 0 background
+        contoured = morphological_chan_vese(predict_area, 100, init_level_set=level_set)
+
+        # contoured area should get original label value
+        contoured_label = np.where(contoured, label, 0)
```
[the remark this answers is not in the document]
geneva-miller: Looked into this as well and you're right, `contoured_label = contoured * label` gives the same result and is faster :+1:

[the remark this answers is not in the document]
willgraf:
```suggestion
        contoured_label = contoured_label * label
```

browser/caliban.py
geneva-miller: I didn't find a difference between these with timeit, but I think this logic is easier to follow

geneva-miller (approved):
Reviewed and user tested on some edge cases, very excited to have annotators start using these actions! :tada:

willgraf (approved):

PR #219, opened 2020-12-23
Add .ebignore to reduce app deployment size.

willgraf:
## What
* Add new `.ebignore` based off of the Python `.gitignore`.

## Why
* The `.ebignore` file prevents `eb deploy` from packaging Python development and build files, reducing the application bundle size and upload/deployment time.


willgraf:
Closing this PR as the `.ebignore` file is added in #235

Issue #220, opened 2021-01-08
Reloading project URL does not switch between features when undoing past actions


tddough98:
**Describe the bug**
When editing labels on different features in browser Label, reloading the page and undoing the actions will not switch between features, making changes on other features occur out of sight.

**To Reproduce**
1. Create a new project in browser Label with multiple features (such as the first example file on the Label homepage '3D organoid annotation').
2. Edit the labels on feature 0 (like painting).
3. Switch to a different feature with 'f'.
4. Edit the labels on the new feature.
5. Reload the page.
6. Undo all the actions. The actions will be correctly undone, but the changes occur on out-of-sight features.

**Expected behavior**
When actions occur on different frames in the stored history on the backend, we should switch to the edited feature before undoing the action so the changes are visible.

PR #221, opened 2021-01-08
Fix action history initialization to change frames correctly

tddough98:
## What
This is a quick bug fix that affects reloading the action history when revisiting a project URL. The constructor for a ChangeFrame action changed between #212 and #214, but did not change in initializeActionHistory, causing the frames to not change while undoing past actions after reloading the page.

One issue highlighted by this fix is that initializeActionHistory does not switch between features when past actions occurred on different features. I made an issue #220 to document this.

## Why
When actions occur on different frames, we should switch to the frame where the action occurs before undoing it so that the changes are visible.

### DIFF
--- a/browser/static/js/history.js
+++ b/browser/static/js/history.js
@@ -115,7 +115,8 @@ class History{
       for (let frame of actionFrames) {
         // Display unedited frame before loading edited frame
         if (frame != prevFrame) {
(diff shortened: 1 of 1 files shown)


willgraf (commented):
Looks like a simple fix. Would it be possible to add a test that makes sure the frames are changing properly?

willgraf (approved):

PR #226, opened 2021-02-12
Rework Label javascript with MVC pattern and statecharts

tddough98:
This PR is on an Elastic Beanstalk instance for testing here: [statechart-test.q2qqtvkafv.us-east-2.elasticbeanstalk.com/](http://statechart-test.q2qqtvkafv.us-east-2.elasticbeanstalk.com/)

## Major Changes:

- MVC pattern to split main_zstack.js
- use XState & statecharts to orchestrate actions
- use webpack to bundle javascript modules

### Model View Controller pattern

Previously, data management, rendering, and event handling were mixed together through out main_zstack.js, making the behavior of the app tightly coupled with many different functionalities. By separating out main_zstack.js into separate modules along a Model View Controller pattern, changes become easier to implement and understand.

In terms of their responsibilities, the Controller intercepts browser events and passes them to the Model (or really passes them to the statechart, which drives the Model). The Model holds and manages the ground truth data. The View renders the Model contents onto the UI.

### Statechart

Statecharts are an extension of state machines that let us to declare behavior in each state. We define states and transitions between states. Statecharts can launch actions when entering a state, exiting a state, or moving between states, realizing side effects like updating data or the UI and allowing the statechart to drive the application. We use [XState](https://xstate.js.org/docs/) to implement our statechart. More background on statecharts can be found here: [statecharts.github.io](https://statecharts.github.io/).

Statecharts have a states running in parallel, where the statechart is in each parallel states simulataneously. These states are useful for orthogonal functionality, where the responsibilities of each parallel state do not affect each other. For example, we have a image adjuster state in parallel that adjusts brightness, contrast, highlighting, and inversion, so these actions are always available independent of the rest of the statechart.

Statecharts can also have hierachical states that refine behavior. We have a toolbar state that determines what mouse clicks do on the canvas. Some tool states are simple and do an action after each click, like autofitting or flooding. Other tools, like painting, thresholding, and watershed, have substates that ensure that the action flow progresses smoothly. For example, the watershed state checks that the first seed selection is not background and the second seed selection is a different point within the same label before launching the action.

XState encourages an [Actor model](https://xstate.js.org/docs/guides/actors.html#actor-api), where the application is structured with actor micro-services that pass messages between each other. The only place this model is used is for a separate statechart that communicates with Label's backend Flask server. The main statechart forwards messages like EDIT, SETFRAME, or UNDO to the backend statechart, which then communicates with the server and sends back LOADING and LOADED messages to the main statechart. The main statechart updates its state based on these LOADING and LOADED messages and knows nothing else about the server requests. A number of states within the main statechart have loading substates that prevent additional server requests from being sent while waiting for results.

### Webpack

Webpack is a JavaScript module bundler. Webpack builds a dependency graph of modules in an application and builds a bundle of all these modules. We instead load a single bundle into our application instead of a mass of dependencies. It also offers benefits like bundle minimization to reduce the size of the javascript assets.

Previously, we loaded all of our javascript files as scripts. All scripts used the same namespace, and we had to load the scripts in a careful order so that dependencies were loaded before the scripts that used them.  Now with Webpack, each file becomes a module that exports only what other modules need to use from it. Each file's interface becomes small and well-defined, instead of every file having access to everything in every other file. Webpack also captures external libraries, making them easier to incorporate into our project.

Webpack can also manage assets like CSS through its plugins, but in this PR, only the javascript files are bundled.

## Minor Changes

- add tabs to dropdown documentation
- add route to toggle RGB mode & a keybind (Command + B) to toggle RGB mode in the browser
- add flood action that combines flood contiguous and fill hole actions
- remove swap all frame action
- remove pixel-only and label-only URL settings
- changed setup in some tests in label_test.py to reshape an array instead of asserting the intended shape


browser/blueprints.py
```diff
@@ -122,6 +122,34 @@ def change_display(token, display_attribute, value):
...
+    current_app.logger.debug('Set RGB to %s for project %s in %s s.',
+                             rgb, token,
+                             timeit.default_timer() - start)
+    return jsonify(payload)
+
+    project = Project.get(token)
+    if not project:
+        return abort(404, description=f'project {token} not found')
+    payload = project.undo()
+
```
willgraf, line 151: Unreachable code?

browser/src/actions.js
```diff
@@ -0,0 +1,368 @@
+/* eslint-disable comma-dangle */
```
willgraf, line 1: You can configure this in the `.eslintrc` for project-wide settings.

browser/src/keydownLookup.js
```diff
@@ -0,0 +1,44 @@
+/* eslint-disable quote-props */
```
willgraf, line 1: I think we can drop this ignore and change the object like:

```javascript
const keydownLookup = {
  a: 'keydown.a',
};
```

browser/src/controller.js
```diff
@@ -0,0 +1,174 @@
...
+    this.model.updateMousePos(this.model.canvas.rawX, this.model.canvas.rawY);
+  }
+
+  /**
+   * Disable scrolling with the wheel (on the canvas) or with keys (everywhere)
+   */
+  overrideScroll() {
+    const canvasElement = document.getElementById('canvas');
+    canvasElement.onwheel = (event) => { event.preventDefault(); };
+    canvasElement.onmousewheel = (event) => { event.preventDefault(); };
```
willgraf, line 64: With single line arrow functions, you can simplify this!

```suggestion
    canvasElement.onwheel = event => event.preventDefault();
    canvasElement.onmousewheel = event => event.preventDefault();
```

browser/src/history.js
```diff
@@ -103,38 +104,6 @@ class History{
...
-  /**
-   * Initializes the action history using the history stored on the backend
-   * @param {*} actionFrames frame of each action performed on the project
-   * @param {*} initFrame frame displayed at the bottom of the undoStack
-   * @param {*} finalFrame frame displayed at the top of the undoStack
-   */
-  initializeHistory(actionFrames, initFrame = 0, finalFrame = current_frame) {
-      // Initialize undoStack to with actions recorded on backend
-      let prevFrame = initFrame;
-      for (let frame of actionFrames) {
```
willgraf, line 115: Maybe I'm missing it elsewhere, but did this get moved or just removed?

browser/src/tool.js
```diff
@@ -0,0 +1,25 @@
...
+  var instances = M.Collapsible.init(elems, {
+    accordion: true,
+    inDuration: 400,
+    outDuration: 400,
+  });
+
+  // Initialize tabs
+  var elems = document.querySelectorAll('.tabs');
+  var instances = M.Tabs.init(elems, {});
+});
```
willgraf, line 21: We can discuss offline, but what are the differences between index.js and tool.js? It looks like we have some duplicated init scripts?

browser/webpack.config.js
```diff
@@ -0,0 +1,16 @@
...
+    tool: './src/tool.js',
+  },
+  output: {
+    filename: '[name].js',
+    path: path.resolve(__dirname, './static/js'),
+    libraryTarget: 'var',
+    library: '[name]'
+  },
+  mode: 'production',
+}
```
willgraf, line 16: When the PR is ready let's have both a dev config and a prod config.

browser/src/model.js
```diff
@@ -0,0 +1,235 @@
...
+        this.maxLabelsMap.set(i, Math.max(...Object.keys(tracks[key]).map(Number)));
+      } else {
+        // if no labels in feature, explicitly set max label to 0
+        this.maxLabelsMap.set(i, 0);
+      }
+    }
+  }
+
+  handlePayload(payload) {
+    if (payload.imgs) {
```
willgraf, line 215: Would it make sense to also use the `Object.prototype.hasOwnProperty` on `payload` checking for property `imgs`?

browser/templates/tool.html
```diff
@@ -71,37 +67,45 @@ <h4>{{ settings['title'] | title }}</h4>
...
+    }
+    function upload() {
+      // disable button and show loading bar
+      document.getElementById('submit').classList.add('disabled');
+      document.getElementById('loading-bar').classList.add('active');
+      const promise = $.ajax({
+        type: 'POST',
+        url: `${document.location.origin}/upload_file/${outputBucket}/${projectID}`,
+        async: true
+      });
```
willgraf, line 92: I liked how you moved more of the HTML scripts into a JS file. Would it make sense to do that here as well?

willgraf (commented):
Wow a lot of changes here! I need to spend some more time on it but I've left a few comments from my first read-through.

browser/blueprints.py
tddough98: Whoops, must have been left behind from copying another route

browser/src/controller.js
tddough98: Fixed here and for other single parameter arrow functions in this module. I noticed that we have mixed parameter names for event handlers `event =>`, `evt =>`, and `e =>`. Should we choose one and stick with it throughout?

browser/templates/tool.html
tddough98: I initially tried moving the whole block to a JS file, but some parts depend on access to the `settings`, which are only accessible through the Flask/Jinja template as far as I'm aware. In particular, we use the settings to access the project ID and the output bucket for uploading to S3.

browser/src/model.js
tddough98: On the Flask side, we set `imgs: false` when there are no images to send, so the payload always has this property.

browser/src/controller.js
willgraf: Yeah, I think it would be nice to have consistent variable naming throughout the project.

browser/src/history.js
tddough98: This was intended to let users to undo edits from past sessions or when reloading the past, but it breaks when editing labels on different features, so I removed it to avoid the bug. The issue is detailed in #220. I'm open to getting it working again, removing it, or leaving it in with the bug. I don't think we have any annotations under way that use multiple features, so I don't think it's a bug that would affect annotation right now

browser/src/backendMachine.js
```diff
@@ -0,0 +1,152 @@
+import { Machine, assign, sendParent } from 'xstate';
+import $ from 'jquery';
+
+const getModel = () => window.model;
+
+const backendMachine = Machine(
```
tddough98, line 6: Screen Shot 2021-02-16 at 2 17 52 PM

Here is a visualization of this state machine.

- this machine is [invoked](https://xstate.js.org/docs/guides/communication.html) by the main deepcell label statechart
- the idle state means that we are not waiting a response & it can receive messages to call specific routes, like edit or undo
- When we receive a message like EDIT or UNDO, we leave the idle state & enter the state to call the routes and wait for a response
- When leaving the idle state, we send a LOADING message back to the parent so the parent can update its behavior
- Once we receive a response, we transiently enter the loaded state to handle the response & send a LOADED response back to the parent

tddough98: This state machine also handles errors from these routes, recording the error and sending an ERROR message, but the app doesn't currently do anything with these messages.

browser/src/statechart.js
```diff
@@ -0,0 +1,615 @@
...
+      trace: [],
+      storedLabel: 0,
+      storedX: 0,
+      storedY: 0,
+    },
+    invoke: [
+      backendMachine,
+    ],
+    states: {
+      rgb: rgbState,
```
tddough98, line 530: Screen Shot 2021-02-16 at 2 18 03 PM

These first states are more straightforward, with most of them simply carrying out actions in response to events.

The rgb state also tracks whether we are viewing an RGB image, and the canvas state tracks whether the space bar is down & we are panning. As these states are [orthogonal or parallel
](https://xstate.js.org/docs/guides/parallel.html), we are in all of these states simultaneously and all the actions are always available (i.e. we can always change the brush size because we are always in the brush state).

browser/src/statechart.js
```diff
@@ -0,0 +1,615 @@
...
+};
+
+const brushState = {
+  on: {
+    'keydown.up': { actions: 'incrementBrushSize' },
+    'keydown.down': { actions: 'decrementBrushSize' }
+  }
+};
+
+const displayState = {
```
tddough98, line 374: Screen Shot 2021-02-16 at 2 18 16 PM

The display state keeps track of whether we are viewing the raw image, the labels, or the overlay.

browser/src/statechart.js
```diff
@@ -0,0 +1,615 @@
...
+        'keydown.z': {
+          target: 'overlay',
+          actions: 'updateCanvas'
+        }
+      }
+    }
+  }
+}
+
+const frameState = {
```
tddough98, line 404: Screen Shot 2021-02-16 at 2 19 40 PM

The frame state lets us to request new images by changing the frame, feature or channel. It includes a loading substate that we transition in/out of on LOADING/LOADED events so we don't request another image while waiting for one.

browser/src/statechart.js
```diff
@@ -0,0 +1,615 @@
...
+    loading: {
+      on: { LOADED: 'idle' }
+    }
+  },
+  on: {
+    SETFRAME: { actions: forwardTo('backend') },
+  }
+};
+
+const confirmState = {
```
tddough98, line 429: Screen Shot 2021-02-16 at 2 17 04 PM

The confirm state manages keybind actions. All of these actions require confirmation with Enter before launching them, or can be aborted with Escape. The view displays a prompt depending on which of the confirm states we are in. This state also has a loading substate so we cannot start an action while waiting for a server response.

browser/src/statechart.js
```diff
@@ -0,0 +1,615 @@
...
+      target: '.watershed',
+      in: '#deepcellLabel.rgb.oneChannel'
+    },
+  }
+};
+
+/**
+ * Handles tool behavior with a loading state.
+ */
+const mouseState = {
```
tddough98, line 315: Screen Shot 2021-02-16 at 2 19 05 PM

The mouse state controls what clicks do on the canvas. Its main state is the toolbar state, which can be in the flood, trim, erodeDilate, autofit, paint, threshold, or watershed substates. It also has a loading state so clicks do not cause extra actions while results are loading. 

The first four tools (shown above) are simple states that launch an action everytime we click.
 


Screen Shot 2021-02-16 at 2 19 27 PM

The paint state and threshold state have substates to enable clicking and dragging.

Screen Shot 2021-02-16 at 4 06 51 PM

The watershed state keeps track of two clicks on the same label.

The threshold, watershed and autofit actions are disabled when viewing an RGB image, so these states transition to the paint state on TOGGLERGB events.

browser/src/statechart.js
```diff
@@ -0,0 +1,615 @@
...
+  initial: 'paint',
+  states: {
+    flood: floodState,
+    trim: trimState,
+    erodeDilate: erodeDilateState,
+    autofit: autofitState,
+    paint: paintState,
+    threshold: thresholdState,
+    watershed: watershedState,
+    hist: { type: 'history' }
```
willgraf, line 288: What does this `type` mean?

browser/src/view.js
```diff
@@ -0,0 +1,511 @@
...
+      prompt = `Predict labels on frame ${frame}?`;
+    } else if (state.matches('confirm.predictAll')) {
+      prompt = 'Predict labels on ALL frames?';
+    } else if (state.matches('confirm.replaceFrame')) {
+      prompt = `Replace label ${secondLabel} with label ${label} on frame ${frame}?`
+    } else if (state.matches('confirm.replaceAll')) {
+      prompt = `Replace label ${secondLabel} with label ${label} on ALL frames?`
+    } else if (state.matches('confirm.swapFrame')) {
+      prompt = `Swap label ${label} with label ${secondLabel} on frame ${frame}?`
+    }
```
willgraf, line 190: Can these states be defined in the state chart? Or is that not recommended?

browser/src/statechart.js
```diff
@@ -0,0 +1,615 @@
...
+  states: {
+    idle: {
+      on: {
+        click: {
+          cond: 'validSeed',
+          target: 'clicked',
+          actions: 'storeClick'
+        }
+      }
+    },
```
willgraf, line 257: What does the `entry` field do? Is it storing the last click even if it's in the `idle` state?

browser/src/statechart.js
```diff
@@ -0,0 +1,615 @@
...
+          cond: (_, event) => !event.shiftKey,
+          target: 'dragging',
+          actions: 'addToTrace'
+        }
+      }
+    },
+    dragging: {
+      on: {
+        mousemove: { actions: 'addToTrace' },
+        mouseup: { actions: 'draw' },
```
willgraf, line 169: In `thresholdState` we use a second object here: `{ target: 'idle' },` which I assume puts the state back into `idle` after the first action is finished.

Why do we use the `target` there but not here?

browser/src/statechart.js
willgraf: So to clarify, the keybinds for both `confirmState` and `frameState` are globally available, but those in the `confirmState` ask for Enter/Esc press and `frameState` does not?

willgraf (commented):
There's a lot going on in the statecharts! It might be a good time to add some tests using [@xstate/test](https://github.com/davidkpiano/xstate/tree/master/packages/xstate-test).

browser/src/backendMachine.js
```diff
@@ -0,0 +1,155 @@
...
+  },
+  {
+    actions: {
+      logError: (context) => console.log(context.error),
+      handlePayload: (context) => getModel().handlePayload(context.data),
+    },
+    services: {
+      edit: (context, event) => fetch(
+        `${document.location.origin}/api/edit/${getModel().projectID}/${event.action}`,
+        { method: 'POST', body: new URLSearchParams(event.args) },
```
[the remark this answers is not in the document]
tddough98: If this promise doesn't resolve for any reason, we transition to the error state, which calls the `logError` action and sends an ERROR event back to the parent. We don't currently do anything with the ERROR event, but we could add behavior on the parent machine when it receives the event.

browser/src/backendMachine.js
```diff
@@ -0,0 +1,155 @@
+import { Machine, assign, sendParent } from 'xstate';
+
+const getModel = () => window.model;
```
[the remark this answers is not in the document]
tddough98: I think I had tried to set up a constant `const model = window.model` to refer to throughout the script, but as the model changes, those changes would not be appear in this constant, so I replaced it with a function. However, `getModel()` is about as long as `window.model`, so I'll put it directly in place throughout this module and in `statechart.js`.

browser/src/statechart.js
tddough98: Hmmm, I think we don't need the transition for either state, as both the `draw` action and the `threshold` action will result in an `EDIT`, `LOADING`, and `LOADED` events that will cause the mouse state to transition in & out of the `loaded` state and then to the toolbar History state to return to `paint.idle` or `threshold.idle`. However, I think it would be better to add in the `{ target: 'idle' }` because we go to the initial or idle state when we use a shallow history, but not a deep history. By always transitioning to `idle`, both the shallow and deep history would behave the same.

tddough98: Just took a second look at the second object `{ target: 'idle' }` on threshold. This is actually a fallback transition when for when we lift the mouse and the threshold bounding box is empty, so we can't threshold. When the box is nonempty, we do the `threshold` action (and don't transition to idle), and otherwise transition back to `idle` so we can draw another bounding box. Based on my previous comment, I think we should add transitions to `idle` for both the paint state and threshold state when we launch the `paint` or `threshold` actions.

browser/src/statechart.js
tddough98: Good catch, that was from a previous workflow with watershed where we entered it on click instead of on a keybind. It would store the coordinates and labels below the mouse when entering the watershed state. These will be overwritten once we click in the idle state and call `storeClick` again, so this doesn't cause any bugs, but we should remove it.

browser/src/statechart.js
tddough98: This creates a (history state)[https://xstate.js.org/docs/guides/history.html]. This allows us to return to using the same tool once the previous action resolves, so we can use the same tool repeatedly.

browser/src/statechart.js
tddough98: The frameState does not edit the labels, it just fetches different image slices from the backend, so there's no need to confirm those actions. The actions in confirm state do edit labels, so we ask the user to confirm the change in case they hit the keybind accidentally.

browser/src/view.js
tddough98: Not sure what you mean by defining these states on the state chart. All of these states (`confirm.predictFrame`, `confirm.predictAll`, etc) are already in the state chart, and we're checking if we are in that state here because we need to display different prompts depending on the state.

I tried storing the prompt on the state machine initially, but that caused some bugs because the prompt should update as the selected labels change, so putting them in the view lets us access the correct selected labels.

browser/src/controller.js
```diff
@@ -0,0 +1,171 @@
...
+    } else if ((event.ctrlKey || event.metaKey) && event.key === 'b') {
+      this.service.send('TOGGLERGB');
+    } else if (keydownLookup[event.key]) {
+      this.service.send(keydownLookup[event.key]);
+    }
+  }
+
+  handleKeyup(event) {
+    if (event.key === ' ') {
+      this.service.send('keyup.space')
```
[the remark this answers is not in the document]
tddough98: This is the only keyup event. We could create another keyupLookup table, but I thought it wasn't worth it for just one case.

browser/src/controller.js
```diff
@@ -0,0 +1,171 @@
...
+   * Handle all keybinds
+   * @param {KeyboardEvent} event
+   */
+  handleKeydown(event) {
+    if ((event.ctrlKey || event.metaKey) && event.shiftKey && (event.key === 'Z' || event.key === 'z')) {
+      this.redo();
+    } else if ((event.ctrlKey || event.metaKey) && (event.key === 'Z' || event.key === 'z')) {
+      this.undo();
+    } else if ((event.ctrlKey || event.metaKey) && event.key === 'b') {
+      this.service.send('TOGGLERGB');
```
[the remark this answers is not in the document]
tddough98: These cases are not in the keydown lookup table because they are checking for multiple keys at once (Ctrl or Command + something else) and they also behave differently, calling a function or sending a named event instead of sending a `keydown.[key]` event. I don't see how we can fold them into the keydown lookup table. Maybe we could have function for each case, but that doesn't sound any better than what we have.

browser/src/view.js
willgraf: Sorry, I meant storing the prompts as part of the state machine, but you've addressed that.

geneva-miller:
Initial thoughts about the user experience:
- it makes me antsy to have the brush (or another action) always ready to go with a click. I would really prefer if there were a "neutral" state just for visualization, where clicks and right clicks change the foreground/background labels but do not make any changes to them (I think both regular clicks and shift-clicks could be allowed in this state). I'm not sure what the best way to access "neutral state" would be: "ESC" from the other action modes? its own keybind? As soon as I saw that "B" was used to toggle to the brush, I tried pressing it to turn the brush off--being able to turn action states "off" and enter "neutral" state would work
- I think I'm okay with panning by holding down the spacebar and moving (rather than clicking and dragging) the mouse. I'm not quite used to it but I think that behavior will help prevent accidental drawing with the brush when panning. I think click events should basically be disabled while spacebar is held down, though. Space + click-and-drag is a common enough way to pan images that I think there's going to be unintentional/unneeded clicking (I know I was doing that)
- selecting labels while in an action mode also sends an action (e.g. in autofit mode--shift-click to select a label to set as foreground also sends the autofit action for that label), but I think using shift-click or shift-right click should never send actions
- highlighting behavior with action states is kind of discordant to me (e.g. in autofit mode, clicking on any label will autofit it but not update the highlights, so there may be two other labels highlighted as foreground/background that have nothing to do with the action just finished). What I'm thinking here is that clicking to apply an action to a label should also set the foreground to that label. I think it would help visually when applying actions, and could also make a sequence of actions smoother (autofit label, it needs a slight touch-up, switch to brush mode and the correct label is already selected). Some action modes (autofit) I think would also benefit from not displaying the second highlight, and maybe not allowing the second highlight/background label to be changed with shift-right click if it isn't displayed
- I'm not quite used to having to explicitly set the "background" value back to background when finished with conversion brush functionality, but I don't have alternative suggestions I feel strongly about. I think the "X" key works nicely to swap foreground/background. I noticed that setting foreground and background to be the same value is easy to do (and they can both be set to the "background" label 0), but I would like to prevent that case.
- I absolutely love the information tabs across the top. Way more user-friendly than what it used to be, looks nice, I found them very helpful when trying out the changes
- it looks like the right-click menu is suppressed for right-clicks (which is what we wanted) but not for shift-right clicks. I think most users will never want to see that menu pop up while they're using the tool, however... I often use the right-click menu to save the canvas image when I'm creating instructions or examples for annotators. This is more of a bonus than a necessary feature, but maybe we could add a function for downloading the canvas image? It would not need to be accessible via a keybind or anything, more of a convenience that I could access through the console
- +/- keys for zooming appear to be swapped

Overall, I'm quite excited about these changes. Statecharts seem really appropriate for handling the different things happening in the app, and I think the revised approaches to the various tools have the potential to make user workflows easier and more intuitive.

browser/package.json
```diff
@@ -1,11 +1,36 @@
...
+    "test": "echo \"Error: no test specified\" && exit 1",
+    "start": "webpack --config webpack.dev.js",
+    "build": "webpack --config webpack.prod.js"
+  },
+  "keywords": [],
+  "author": "The Van Valen Lab",
+  "license": "SEE LICENSE IN ../LICENSE",
   "devDependencies": {
+    "@xstate/inspect": "^0.4.1",
     "eslint": "^6.8.0",
```
willgraf, line 16: We can also want to declare other dependencies here (CSS and JS files).

willgraf (changes requested):
Let's discuss tomorrow but I left a few comments.

I think there should be more error handling with the `fetch` calls. They don't throw an error if they get a 500 code, so we should check the response code when we get it.

browser/static/css/main.css
```diff
@@ -97,3 +94,22 @@ main {
...
+  font-size: 0.85em;
+  font-weight: 700;
+  line-height: 1;
+  padding: 2px 4px;
+  white-space: nowrap;
+}
+
+#error {
+  color: red;
+}
```
willgraf, line 115:
```suggestion
}

```

willgraf (approved):

PR #236, opened 2021-04-14
Bebi 205 documentation

tddough98:
## What

Fixes some typos & adds some excerpts from previous workshop material to the dropdown documentation.

Two functional changes:

- prevent default behavior for Ctrl+B (which is a bookmark shortcut in Firefox)
- rerender the canvas on state transitions so the brush appears immediately when switching to the brush tool

## Why

Trying to get in some last-minute polish before the Label workshop later today.

### DIFF
--- a/README.md
+++ b/README.md
@@ -143,7 +143,7 @@ Press <kbd>Shift</kbd> + <kbd>F</kbd> to view the previous feature.
 
 ### Adjusting the Canvas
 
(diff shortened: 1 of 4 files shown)


tddough98:
Fixed another bug where increment channel or feature actually decremented

willgraf (approved):
LGTM.

PR #238, opened 2021-05-06
Temp convert npz column

tddough98:
## What
We introduce a new custom column type that reads pickled image data in the database and writes back to the database as a compressed npz. We also add a script to go through all projects in the database and convert pickled data to npz.

## Why
The deepcell_label database has been filling up very rapidly over the past months since we've been tracking the history of each project. The footprint for a compressed npz in the database can be up to 100x smaller than a pickled object, greatly reducing the cost of storing this data.

The temporary conversion column type and the conversion script allows to continue to use the existing database while we we reformat this data. Once the conversion script has completed, we can swap out the PickleToNpz column to a simple Npz column as all the data moving in and out of the database will already be in npz format

### DIFF
--- a/deepcell_label/models.py
+++ b/deepcell_label/models.py
@@ -10,6 +10,8 @@ import logging
 import os
 import timeit
 from secrets import token_urlsafe
+import pickle
+import io
 
 from flask import current_app
 from flask_sqlalchemy import SQLAlchemy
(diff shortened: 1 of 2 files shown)


tddough98:
Test coverage only drops because of adding the pickle_to_npz.py script. We'll only keep this script in the repo temporarily, so I don't think we need to write tests for it.

PR #242, opened 2021-05-12
Switch PickleToNpz for Npz column and remove conversion script

tddough98:
Here, we remove the script used to convert pickled image data to npz and replace the `PickleToNpz` column with a `Npz` column. As the conversion script has been run, we expect all image data stored within the database to be compressed npz, not pickled objects.


willgraf (approved):
This PR closes the door on the database migration and allows us to move forward saving data to the DB with a minimal footprint. 🎉

PR #248, opened 2021-06-22
Rework DeepCell Label with React

### LINKED ISSUE #11: Faster scroll through images option

### LINKED ISSUE #52: edit mode visual improvements
Miscellaneous small improvements would give the pixel-editing mode a nicer feel. These include:

- [ ] color of brush preview should match color of cell that is being annotated

- [ ] brush should be somewhat transparent

- [ ] annotations currently are composited on top of image, making them difficult to see against dim background. Maybe some combination of compositing and adjusting alpha?

- [ ] (bonus) it would be awesome if the transparency of brush/annotations could be adjusted by the user. the simplest version of this could be toggling the overlay on/off (this is implemented in desktop caliban) but a nicer version would perhaps include a slider that corresponds to the alpha value when drawing the overlays

### LINKED ISSUE #96: disable unused channels
Is there a way that we can disable interactions with channels that don't contain data? For example, so that the brightness can't be adjusted, empty channels don't need to be cycled through to get to actual channels, and their values don't show up on the brightness settings? 

I think this would be useful to reduce confusion for annotators.

### LINKED ISSUE #97: grey out text when waiting for keyboard response
It's sometimes hard to notice that caliban is waiting for you to hit s for save, or space to confirm, for a specific action that you've selected. It would be a nice visual queue for this if the descriptive text, such as channel, label, etc, would grey out when awaiting a response to more quickly guide the eye towards the bottom of the screen

### LINKED ISSUE #99: new border segmentation has different opacity when being created
When in border segmentation view, the overlay that is being created has a different transparency setting than the already created segmented view. 

For example, if adding more area to an existing cell, when drawing in pixel editing mode, the parts of the cell that appear underneath the newly added area, before the brush is released, has a different look than the areas that have already been created. 

I'm not sure if this is a side-effect or intended, but it makes difficult to determine where to stop the segmentation contour if the same brightness value appears differently under the brush than outside of it

### LINKED ISSUE #102: remove small holes from label action
Alternative to hole fill action. Sometimes small holes within a label are particularly annoying to fill by clicking after hole fill prompt. This action could be a variation on hole fill (f for hole fill -> new keybind instead of clicking on pixel -> holes within that label are filled). This would use skimage.morphology.remove_small_holes instead of flood_fill.

Use case: annotations that have one or more very small holes inside. Size of hole to fill would likely be set at <10.

### LINKED ISSUE #110: Browser improvements to make annotation easier

### LINKED ISSUE #115: color adjustment scroll wheel requires multiple scrolls
To adjust the colors in RGB mode, the scrollwheel allows the user to increase or decrease the brightness. However, if you move the scrollwheel forward a single click, nothing changes. It looks like the "scrollwheel" doesn't activate unless multiple scrolls are input, which makes it tricky to adjust MIBI/IMC data which has small quantized values.

### LINKED ISSUE #131: Dynamic resizing of HTML for image
I spoke with Geneva about this already, neither of us was sure what the best way to approach this would be. 

Currently, only a small portion of the screen is actually showing the imaging data. If there was a way so that the image could be more intelligently resized, I think this would increase the speed of annotations significantly, and make them more accurate, especially in tissue. 

image

### LINKED ISSUE #132: Remove unnecessary display text for RGB
Since RGB mode right now it being used just for 2D data, it would make the UI much cleaner if we removed the extra information that's only relevant for 3D data. 

In particular, I think all of these can be removed:
Frame, channel, feature, highlight cells, highlighted cells, slices

### LINKED ISSUE #140: Deeper refactor of image display in browser
Currently, image modifications are made each time an image is displayed. For example, moving the brush across the canvas triggers a redraw because the brush preview must be updated, but this also causes a redraw of the raw image, followed by greyscaling it (possibly contrast + invert as well), compositing the labels on top, and applying other changes, such as outlines. 

Instead, modified images should be stored in memory at a few different stages to reduce the number of redundant image modification steps. (This can follow pattern of adjusted_seg image in zstack js: the source of an Image object is updated from a hidden canvas when the highlight is changed.) Images may include:

- contrast/brightness adjusted raw image

- solid-color label image with highlight applied

- raw image + label image composited together

- raw image + label image with outlines and/or translucent highlighting applied on top

Each of these images can be prepared and saved from the same hidden canvas. Ideally, image updates would trigger upon changing (setting) relevant values, eg highlighted labels, contrast, etc.

### LINKED ISSUE #155: store filter settings per channel
Desktop version stores brightness adjustments on a per-channel basis. Filter settings (adaptive hist eq, sobel filter, etc) should also be stored on a per-channel basis. This is particularly useful for files that contain both fluorescent and phase channels.

### LINKED ISSUE #158: handle_draw sends duplicate mouse trace coordinates in payload.

### LINKED ISSUE #169: Look into compressed npzs
Using `numpy.savez_compressed` will result in smaller file sizes than `numpy.savez`, which is currently what we use to create our npz files. We should look into using compressed npzs, which would need to be compatible with both caliban and the data pipeline.

### LINKED ISSUE #171: Browser caliban can't handle (large?) floats in channel data
I tracked down the issue with the NPZs not loading, it looks like it was due to the dtype of the X data. My guess is the combination of a 1024x1024x image and the fact that the data was float64 lead to a memory error, but I'm not 100% sure. 

Here's a 1024x1024 image, int16, that loads fine: https://caliban.deepcell.org/caliban-input__caliban-output__multiplex__20200413_Human_Agreement__Roshan_P101__output__correction_int__stitched_data_int.npz?pixel_only=False&label_only=False&rgb=True

Here's the same 1024x1024 image, but with float64, that fails to load: 

https://caliban.deepcell.org/caliban-input__caliban-output__multiplex__20200413_Human_Agreement__Roshan_P101__output____correction__stitched_data.npz?pixel_only=False&label_only=False&rgb=True

Please don't modify the data in the first link, it's currently being worked on for the human comparison analysis.

### LINKED ISSUE #203: fully replacing label with conv brush doesn't clear it from cell_info

### LINKED ISSUE #220: Reloading project URL does not switch between features when undoing past actions

### LINKED ISSUE #234: misc issues introduced with MVC rework
**Describe the bug**
Typos in infopane
`Press the W W to use the Watershed tool.`
```
feature tells us which labeled features we are viewing
Features may be [FILL IN] 
```

Cycling through foreground values also changes background values (undesired behavior).

**To Reproduce**
1. Select a label (eg set 1 as the foreground and background as the background)
2. Press `[` to cycle backwards through labels (or `]` to cycle forwards
3. When labels are cycled around to the first or last label in the file, the background label changes

**Expected behavior**
When cycling between foreground labels, the background value should remain unchanged, even when changing between first/last label.

**Additional context**
There may also be other unintended or undesired behaviors that I haven't found yet via user testing

### LINKED ISSUE #244: Include a link to the GitHub in the Footer
We could include a link to the GitHub issues page in the footer of the page (much like on deepcell.org). Since we expect some new users to use the Visualizer, a link to the GitHub issues page could help funnel any issues or feature requests to the right place.

tddough98:
This PR provides a complete rework of user interface for DeepCell Label. We use [React](https://reactjs.org/) to build the UI. This PR also build upon previous work with [XState](https://xstate.js.org/docs/) and integrates it with React so that state is managed by actors instantiated by XState and React components use hooks to access and render that state into UI elements.

We also include some changes to the backend to better serve the React-based frontend, namely changed 

## React components

We bootstrapped this React app with create-react-app, so this PR contains the tooling and dependencies installed by CRA.

The UI is made up of Controls on the left, and a Canvas that fills the available space to the right.

### Canvas

The Canvas folder contains everything for the interactive canvas. The top-level Canvas component manages styling and event handling, the RawCanvas folder handles adjusting and showing the raw images, the LabeledCanvas folder handles overlaying the labels on the raw image, and the ToolCanvas folder has components to show tools like brush or threshold.

`canvasUtils` has functions to manipulate [ImageData](https://developer.mozilla.org/en-US/docs/Web/API/ImageData) objects.

### Controls

The ImageControls folder has the controls on the far left used that control what and how images are shown. There is a SubmitButton on top, then a FrameSlider, then the LabelControls, then the RawControls.

The LabelControls include a FeatureSelect, a OutlineToggle, and a OpacitySlider.
The RawControls include a ColorModeToggle to switch between viewing one or multiple channels, then either GrayscaleControls when viewing one channel or RGBControls when viewing multiple channels.
The GrayscaleControls have a switch to invert the channel and sliders for range, brightness, and contrast, and the RGBControls have a LayerController for each channel that controls the layer's color, its range, and whether the layer is shown.

The Toolbar folder has the controls just to the left of the canvas that manage how we edit the labels (like picking a tool, selecting labels.
The UndoRedoButtons are on top, then the Toolbar, then the ActionButtons (swap/replace), and finally a palette that shows if there is a foreground/background selected.

### Other components

Other simple components include the Navbar, the Footer, and the Instructions.

## State management with XState

To provide the state machines to components throught the React, we define a top level ServiceContext component that wraps the whole application. In this component with define a [Context](https://reactjs.org/docs/context.html) that provides the top level state machine to any child of ServiceContext. These children can access the state through hooks defined in this module. At a high level, we can image the React component tree and the XState actor hierarchy as two separate trees that touch each other through these hooks.

Components subscribe to some part of these state machines and rerender when the state updates. 

### Actor hierarchy

The root actor is still `deepCellLabelMachine`, which fetches initial project data, and then spawns `canvas`, `image`, `tool`, `undo`, and `api`.

`canvas` tracks where the canvas is within the image, such as zoom, position, and mouse
position. It also manages mouse events and forwards them to `tool`.

`image` tracks the current frame, and spawns `labeled` and `raw`. `feature` manages the current segmentation, and spawns a feature machine to fetch the data for each segmentation. `raw` manages the current raw image, and spawns a channel machine to fetch the data for each channel. `raw` also spawns two "color modes" `grayscale` and `color` and passes them all the channel actors. The `grayscale` color mode tracks which single channel is displayed in grayscale mode. The `color` color mode spawns a layer for each displayed channel and each layer manages its own image adjustment settings for that channel.

`tool` tracks the current tool we're using to interact with the labels. The `tool` will spawns an actor for the current tool that has custom behavior for mouse events sent from `canvas`. The `tool` actor also tracks the selected labels (the foreground and the background).

`api` handles requests to the Label backend that edit the labels. The events that can edit labels include `EDIT`,  `BACKEND_UNDO`, and `BACKEND_REDO`. Once `api` gets a response from the backend, it sends a `EDITED` event to the `labeled` machine to reload the data that changed.

`undo` can manage any machine that responds to `SAVE` and `RESTORE` events. When `undo` receives an `ADD_ACTOR` event, it creates a `history` machine for that actor that will send a `SAVE` event to the actor when we edit the labeling, and the actor must respond with a `RESTORE` event that contains the state that the actor should restore on undo/redo. Later, when `undo` is undoing or redoing an edit, the `history` machine will send that `RESTORE` event back to the actor and the actor is responsible for restoring their state based on this event.

### Hotkeys

Hotkeys are implemented as [invoked callbacks](https://xstate.js.org/docs/guides/communication.html#invoking-callbacks), where the hotkey listeners are setup when entering the states/machines that use the hotkeys and are torn down on exit.

## Project management

We've added prettier to the project, an opinionated code formatter than enforces consistent style. Prettier formats javascript, css, html and other browser focused formats, complementing out existing linting for python. We also use husky and lint-staged to add a pre-commit hook so any code committed to the repo will be formatted with prettier.

Typescript has been added as a dependency, but we have no types defined yet. Adding typescript now improves the intellisense for the project in VSCode and will make it easier to add types later.

## Changes to Flask backend
The backend changes include:

- a new URL based approach to creating projects, which includes
  - loading raw assets and labeled assets separately
  - support for loading from ZIPs, where separate files in a zip are treated as different channels or features
  - reshaping input arrays based off an 'axes' string like 'ZYXC'
- routes to serve project data separately, rather than in a bundled JSON, including

Here is a table of the new routes, what data they serve, and the data format:

| Route                              | data served                                                                                                                        | data format     |
|------------------------------------|------------------------------------------------------------------------------------------------------------------------------------|-----------------|
| `/api/project`                     | project metadata, like image dimensions                                                                                            | JSON            |
| `/api/raw`                         | raw image slices                                                                                                                   | grayscale PNG   |
| `/api/labeled`                     | labeled image slices                                                                                                               | color PNG       |
| `/api/array`                       | array of label at each pixel, with label borders in negative                                                                       | compressed JSON |
| `/api/instances`                   | label metadata<br>at the moment, only the frames each label is present in<br>used on frontend to know which labels in each feature | JSON            |

## Issues

@willgraf has reviewed outstanding issues and this PR closes these issues

Fixes #11
Fixes #52
Fixes #96
Fixes #97
Fixes #99
Fixes #102
Fixes #110
Fixes #115
Fixes #131
Fixes #132
Fixes #140
Fixes #155
Fixes #158
Fixes #169
Fixes #171
Fixes #203
Fixes #234
Fixes #244
Fixes #220

There are a few known issues introduce by this PR 

### Multichannel display with more than 3 channels #256 

Loading and displaying channels beyond the initial 3 channels does not work properly. Adjusting these higher channels seems to redraw the without clearing the canvas, filling the canvas with a single color. Switching between the higher channels also does not seem to work properly. 

This is likely an issue with staging channels for display when switching frames and channels. The intended behavior is for a frame to render as soon as all displayed channels are loaded for the next frame, and for a channel to render as soon as it is loaded for the current frame. As either the displayed channels or the current frame can change while waiting for the other to load, we track the loading frame or loading channels separately from the current frame/channels.

### Undo lifecycle management

Currently, we add an action to the undo history as soon as we send the action, rather than when we get a response from the action on the backend. As a result,

- sending an action before the last action responds add an extra element in the undo history, which would let us undo an action that never happened
- undoing an action before the last undo responds removes an extra element in the undo history, only undoing the first action

### Backend undo implementation

Actions that don't change the labeling are stored in the action history. Instead, we should only record actions that actually change the data so undoing always shows a visible difference in labels. We will also need to update the frontend undo to only add an element to the history once the backend responds that the labels have changed.

Actions don't record which feature they edit. Currently, we compute which feature changed. We should move towards recording that data directly with the action.

Each action has constant time redo, but undo scales with the number of actions as undo has to search through previous actions to find the last version from of a frame. For example, if a frame is edited 100 actions apart, then for undo, we have to look back to the action 100 frames ago, but redoing the action is stored directly with the action.


deepcell_label/blueprints.py
```diff
@@ -102,6 +103,22 @@ def semantic_instance_labels(project_id, feature):
...
+@bp.route('/api/colormap/<project_id>/<int:feature>')
+def colormap(project_id, feature):
+    project = Project.get(project_id)
+    if not project:
+        return jsonify({'error': f'project {project_id} not found'}), 404
+    max_label = project.get_max_label(feature)
+    colormap = matplotlib.pyplot.get_cmap('viridis', max_label)
+    colors = map(matplotlib.colors.rgb2hex, colormap.colors)
+    # Add no label and new label colors
+    colors = ['#000000'] + list(colors) + ['#FFFFFF']
```
willgraf, line 115: Prefer to insert values into the colors list than to concatenate. As written this is creating 4 lists when we only want 1.

willgraf (changes requested):

visualizer/src/Controls/ImageControls/ImageControls.js
```diff
@@ -0,0 +1,53 @@
...
+      {raw && <RawControls />}
+    </Box>
+  );
+};
+
+// {/* <TableRow className={styles.row}> */}
+// {/* <TableCell className={styles.cell}> */}
+// </TableCell>
+// </TableRow>
+
```
willgraf, line 52: let's just clear this out.

visualizer/src/Controls/ImageControls/LabeledControls/OpacitySlider.js
```diff
@@ -0,0 +1,57 @@
...
+
+function OpacitySlider() {
+  const labeled = useLabeled();
+  const opacity = useSelector(labeled, state => state.context.opacity);
+
+  const handleOpacityChange = (event, newValue) =>
+    labeled.send({ type: 'SET_OPACITY', opacity: Number(newValue) });
+
+  const handleDoubleClick = event =>
+    labeled.send({ type: 'SET_OPACITY', opacity: 0.3 });
```
willgraf, line 28: Double-clicking returns opacity to the default value? Why 0.3?

visualizer/src/Controls/ImageControls/LabeledControls/OutlineToggle.js
```diff
@@ -0,0 +1,47 @@
...
+function OutlineToggle() {
+  const labeled = useLabeled();
+  const outline = useSelector(labeled, state => state.context.outline);
+  const { send } = labeled;
+
+  // Adds mousetrap class so hotkeys work after using switch
+  const inputRef = useRef();
+  useEffect(() => {
+    const input = inputRef.current;
+    input.className = input.className + ' mousetrap';
```
willgraf, line 18: Prefer the string formatting over concatenation:

```suggestion
    input.className = `${input.className}  mousetrap`;
```

visualizer/src/Controls/ImageControls/RawControls/ColorModeToggle.js
```diff
@@ -0,0 +1,63 @@
...
+  const image = useImage();
+  const grayscale = useSelector(image, state => state.context.grayscale);
+  const raw = useRaw();
+  const { send } = raw;
+
+  // Adds mousetrap class so hotkeys work after using switch
+  const inputRef = useRef();
+  useEffect(() => {
+    const input = inputRef.current;
+    input.className = input.className + ' mousetrap';
```
willgraf, line 20: Prefer the string formatting over concatenation:

```suggestion
    input.className = `${input.className}  mousetrap`;
```

visualizer/src/Controls/ImageControls/RawControls/ColorPalette.js
```diff
@@ -0,0 +1,64 @@
...
+    padding: '3px',
+    width: '16px',
+    height: '16px',
+  },
+  icon: {
+    width: '17px',
+    height: '17px',
+    // border: '1px solid black',
+    // borderRadius: '50%'
+  },
```
willgraf, line 38: If possible it would be nice to make this related to the theme instead of using the pixel sizes directly.

I think `em` units are better for responsive web pages as well.

visualizer/src/Controls/ImageControls/RawControls/GrayscaleControls.js
```diff
@@ -0,0 +1,216 @@
...
+}));
+
+const InvertToggle = ({ channel }) => {
+  const invert = useSelector(channel, state => state.context.invert);
+
+  // Adds mousetrap class so hotkeys work after using switch
+  const inputRef = useRef();
+  useEffect(() => {
+    const input = inputRef.current;
+    input.className = input.className + ' mousetrap';
```
willgraf, line 26: Prefer the string formatting over concatenation:

```suggestion
    input.className = `${input.className}  mousetrap`;
```

visualizer/src/Controls/ImageControls/RawControls/LayerOptions.js
```diff
@@ -0,0 +1,80 @@
...
+    '&:hover': {
+      backgroundColor: 'transparent',
+    },
+    paddingLeft: '2px',
+    paddingRight: '2px',
+  },
+}));
+
+function LayerOptions({ layer }) {
+  const [open, toggle] = useReducer(v => !v, false);
```
willgraf, line 36: Can you briefly comment on what this does? I'm not familiar with the reducer hook.

visualizer/src/Controls/Toolbar/UndoRedo.js
```diff
@@ -0,0 +1,86 @@
...
+
+// for adding tooltip to disabled buttons
+// from https://stackoverflow.com/questions/61115913
+const Button = withStyles({
+  root: {
+    padding: 4,
+    '&.Mui-disabled': {
+      pointerEvents: 'auto',
+    },
+  },
```
willgraf, line 24: I notice this is copy/pasted a few times. Any way we can import this instead? Or include it in a global/theme or CSS?

willgraf (changes requested):
This looks great!! I still need to dig more into the state machines, but I went through the react code and left some notes.

Generally, I love it. One thing in particular that caught my eye is the use of `px` values in the styles. If possible I'd like to migrate some of those either into the theme, or into `em` values instead.

visualizer/src/Controls/ImageControls/LabeledControls/OpacitySlider.js
tddough98: 0.3 was the value we used up through the 0.1 release, so keeping it the same makes things look more similar. If the user changes the opacity slider to a different value, then cycling raw -> overlay -> labels with the Z hotkey will use that value instead for the overlay.

visualizer/src/Controls/ImageControls/RawControls/LayerOptions.js
tddough98: The signature for `useReducer` is 
```const [state, dispatch] = useReducer(reducer, initialArg, init);```
so the `open` is the current state and `toggle` is a dispatch method with signature `(state, action) => newState` to update the state based on the current state and an action.

actions in reducers correspond to events for state machines, although there are no actions/events defined here, the reducer always does the same thing when it's dispatch method is called

You can think of a reducer as a state machine with just one state, i.e. the reducer always has the same behavior for an incoming events. State machines build on that model by allowing us to declare states that have different behaviors for each event and can transition between the states on the events. Compare `const [state, dispatch] = useReducer(reducer, initialState)` to `const [currentState, send] = useMachine(machine)`

When `toggle` is called, we take the current value of `open` and flip it. `open` starts as `false`, so the component is closed until it's clicked on

File: visualizer/src/service/service.js [hunk 10]
```diff
@@ -0,0 +1,13 @@
+import { interpret } from 'xstate';
+import createDeepcellLabelMachine from './deepcellLabelMachine';
+
+const location = window.location;
+const search = new URLSearchParams(location.search);
+const projectId = search.get('projectId');
+const bucket = search.has('bucket') ? search.has('bucket') : 'caliban-output';
```
geneva-miller, line 7:
```suggestion
const bucket = search.has('bucket') ? search.get('bucket') : 'caliban-output';
```
We ran into a small bug where setting an output bucket with the url argument isn't working correctly (although submitting a project to the default bucket works fine). It looks like the output bucket gets set to `true` instead of the specified bucket value.

willgraf: Looks like a good find, but I'll let @tddough98 confirm this and commit in the fix.

tddough98: Fix and redeployed with the fix

willgraf:
After reviewing the open issues, I think that this PR will close the following:

Fixes #11
Fixes #52
Fixes #96
Fixes #97
Fixes #99
Fixes #102
Fixes #110
Fixes #115
Fixes #131
Fixes #132
Fixes #140
Fixes #155
Fixes #158
Fixes #169
Fixes #171
Fixes #203
Fixes #234
Fixes #244
Fixes #220

I believe the body of the PR will need to be updated (copy and paste above) in order have the issues be automatically closed.

tddough98:
Coverage has dropped dramatically `Coverage decreased (-28.6%) to 47.71%` because we are including javascript files in the coverage for the first time. We'll have to add more tests in future releases to bump this coverage up again, but I don't think we'll be able to increase coverage in this PR

deepcell_label/url_loaders.py
```diff
@@ -258,7 +258,11 @@ def load_zip(data):
     Treats separate TIFFs as channels.
     """
     zip_file = zipfile.ZipFile(io.BytesIO(data), 'r')
-    channels = [load_tiff(zip_file.open(info).read()) for info in zip_file.filelist]
+    channels = [
+        load_tiff(zip_file.open(item).read())
+        for item in zip_file.filelist
+        if not str(item.filename).startswith('__MACOSX/') and is_tiff(str(item.filename))
```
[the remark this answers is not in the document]
tddough98: Let's move this to `config.py` once we understand what files we might need to exclude more specifically.

visualizer/src/Controls/ImageControls/RawControls/ColorPalette.js
tddough98: This has been changed for our own components like the `SelectedPalette`. This component `ColorPalette` was adapted from another project https://github.com/hms-dbmi/viv, so I'll leave the specific pixel values for now

visualizer/src/Controls/ImageControls/RawControls/LayerOptions.js
```diff
@@ -0,0 +1,75 @@
...
+import Paper from '@material-ui/core/Paper';
+import Popper from '@material-ui/core/Popper';
+import { makeStyles } from '@material-ui/core/styles';
+import MoreVertIcon from '@material-ui/icons/MoreVert';
+import { useSelector } from '@xstate/react';
+import React, { useReducer, useRef } from 'react';
+import { useRaw } from '../../../ProjectContext';
+import ColorPalette from './ColorPalette';
+
+const useStyles = makeStyles(() => ({
```
willgraf, line 15: This CSS was from the viv project?

visualizer/src/Controls/ImageControls/RawControls/LayerController.js
```diff
@@ -0,0 +1,119 @@
...
+      onChange={onChange}
+      onDoubleClick={onDoubleClick}
+      valueLabelDisplay='off'
+      min={0}
+      max={255}
+      step={1}
+      orientation='horizontal'
+      style={{
+        color: color,
+        marginTop: '7px',
```
willgraf, line 77: Here's another px value.

willgraf (commented):
This is great! Clearly, a lot of work went into this, the code looks great and is very readable. We've gone over the code together a few times, and it looks to be in great shape. I gave it one last go-over, but please LMK if I've missed anything you'd like me to take a closer look at.

I love the button pattern you have set up here, easy to read and understand. One idea (for a future PR) is to move the hotkey definitions into a single place, and import them in each Button. This would allow all the hotkey configuration in a single place, without needing to look through each button file. (This is a pattern I've seen commonly used in Java projects).

Additionally, I've flagged a couple of remaining spots where we are using absolute pixel values for CSS. These don't _need_ to be changed, but it's a little pet peeve of mine so please justify their use.

visualizer/src/Controls/ImageControls/RawControls/LayerOptions.js
tddough98: Yes, and there's a `/** Modified from https://github.com/hms-dbmi/viv */` comment at the top of the file to attribute it

visualizer/src/Controls/ImageControls/RawControls/LayerController.js
tddough98: Also from viv

visualizer/src/Controls/ImageControls/LabeledControls/OpacitySlider.js
```diff
@@ -0,0 +1,62 @@
...
+import { bind, unbind } from 'mousetrap';
+import React, { useEffect } from 'react';
+import { useLabeled } from '../../../ProjectContext';
+
+const useStyles = makeStyles(theme => ({
+  opacity: {
+    display: 'flex',
+    flexDirection: 'row',
+    justifyContent: 'space-between',
+    gap: '20px',
```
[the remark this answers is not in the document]
tddough98: Updated

willgraf (approved):
This looks great, very excited to get this merged in 🚀 

Before we do, can you please comment on the Pillow version issue?

tddough98:
Whoops didn't see your last comment before hitting merge! The pillow version was pinned because the last pillow release 8.3.2 had issues with the wheels available for linux installation needed to deploy onto Elastic Beanstalk. Looks like this is fixed now, as this issue shows the original issue and has been solved https://github.com/python-pillow/Pillow/issues/5701. We should unpin pillow the next time we merge into master.

PR #249, opened 2021-06-30
Add an .editorconfig file to standardize file types.

willgraf:
## What
* Add an `.editorconfig` file with standards set for JS and Python.

## Why
* Standardize some editor configuration settings, namely indent-spacing.


tddough98:
Looks good

I was wondering how this played with the addition of Prettier in #248 and it seems like they are a good choice together based on [this post](https://blog.theodo.com/2019/08/why-you-should-use-eslint-prettier-and-editorconfig-together/)

PR #259, opened 2021-08-30
Merge refactors and improvements from tracking branches

tddough98:
As we've continued development on tracking in DCL, we've tweaked the code base to better support the tracking development. Many of the tweaks are more general and can improve the 0.2 release.

The `toolbar` machine previously managed many features, including both what labels are selected, which segmentation tools is active, and interpreting interactions with the canvas. Each of these has been factored out and moved into other or new machines. `selectMachine` manages what labels are selected. `canvasMachine` interprets interactions with the canvas. `segmentMachine` manages what segmentation tool is used.

Keybind control has been moved from machines to components. The components have the tooltips that describe the keybinds and are the UI elements that update in response to the keybinds. This change lets us reuse keybinds to have different functions in different labeling modes, like segmentation and tracking, and better supports multiple projects so that the keybinds for only the current project are active.

There are also some refactors and bug fixes on the backend that will support tracking features or quality control features in in upcoming work.


deepcell_label/url_loaders.py
```diff
@@ -39,94 +39,107 @@ def __init__(self, url_form):
         self.input_axes = url_form['axes'] if 'axes' in url_form else DCL_AXES
         self.output_axes = DCL_AXES
 
+        self.raw_array = None
+        self.label_array = None
+        self.cell_info = None
+        self.cell_ids = None
+
+        self.load()
```
willgraf, line 47: `load()` is still called in the constructor

deepcell_label/blueprints.py
```diff
@@ -246,6 +228,7 @@ def create_project_from_url():
     start = timeit.default_timer()
     url_form = request.form
     loader = url_loaders.Loader(url_form)
+    loader.load()
```
willgraf, line 231: I think this is unnecessary, as `Loader` still calls this in the constructor.

deepcell_label/label.py
```diff
@@ -666,30 +680,6 @@ def action_new_track(self, label):
...
-
-            track_2['parent'] = label_1
-
-            if track_1['frame_div'] is None:
-                track_1['frame_div'] = first_frame_daughter
-            else:
-                track_1['frame_div'] = min(track_1['frame_div'], first_frame_daughter)
-
-            self.labels_changed = True
-
```
willgraf, line 692: Is this just never used?

deepcell_label/labelmaker.py
```diff
@@ -63,55 +60,61 @@ def compute_feature_ids(self, feature):
...
+                          'slices': ''}
             for frame in range(self.num_frames):
-                if cell in feature_labels[frame, ...]:
-                    feature_info[cell]['frames'].append(int(frame))
-        self.cell_ids[feature] = feature_cells
-        self.cell_info[feature] = feature_info
+                if cell in labels[frame, ...]:
+                    info[cell]['frames'].append(int(frame))
+        self.cell_ids[feature] = cells
+        self.cell_info[feature] = info
```
willgraf, line 95: I'm a little confused between `cell_info` and `_cell_info`. I think this is just setting the `_cell_info` dict directly? If I am understanding, can we just update to set `_cell_info` instead to be more clear?

deepcell_label/labelmaker_test.py
```diff
@@ -85,7 +85,28 @@ def test_tracking_one_cell(self):
...
+    def test_tracking_two_features_no_labels(self):
+        labels = np.zeros((1, 1, 1, 2))
+        labeler = LabelInfoMaker(labels, tracking=True)
+        expected_ids = {0: np.array([]), 1: np.array([])}
+        expected_info = {0: {}, 1: {}}
+
+        assert self.compare_cell_ids(labeler.cell_ids, expected_ids)
+        assert labeler.cell_info == expected_info
+
+    def test_tracking_two_features_one_label(self):
```
willgraf, line 97: Probably should add a two labels test since we allow 1+ features now.

deepcell_label/url_loaders_test.py
```diff
@@ -143,7 +143,7 @@ def test_load_raw_png():
 def test_load_raw_zip():
     """Load raw array from zip of tiff files."""
     expected_raw = np.array([1, 2]).reshape((1, 1, 1, 2))
-    expected_labeled = np.array([0, 0]).reshape((1, 1, 1, 2))
+    expected_labeled = np.array([0]).reshape((1, 1, 1, 1))
```
willgraf, line 146: Could you briefly comment on why this is changed?

deepcell_label/url_loaders.py
```diff
@@ -39,94 +39,107 @@ def __init__(self, url_form):
...
             label_array = load_png(data)
-        elif is_tiff(url):
+        elif is_tiff(self.labeled_url):
             label_array = load_tiff(data)
-        elif is_zip(url):
+        elif is_zip(self.labeled_url):
             label_array = load_zip(data)
         else:
-            ext = pathlib.Path(url).suffix
+            ext = pathlib.Path(self.url).suffix
```
willgraf, line 140:
```suggestion
            ext = pathlib.Path(self.labeled_url).suffix
```

willgraf (changes requested):
I love the changes, especially moving the hotkey listening from the machines to the components, much cleaner!

Please review some of the backend changes to make sure they are not overly tracking-oriented.

deepcell_label/label.py
tddough98: This is a legacy action that not used anywhere in the segmentation or tracking tools

deepcell_label/models.py
```diff
@@ -472,29 +476,6 @@ def __init__(self):
...
-        for _, feature in cell_info.items():
-            for _, label in feature.items():
-                slices = list(map(list, consecutive(label['frames'])))
-                slices = '[' + ', '.join(["{}".format(a[0])
-                                          if len(a) == 1 else "{}-{}".format(a[0], a[-1])
-                                          for a in slices]) + ']'
-                label['slices'] = str(slices)
-
-        return cell_info
-
```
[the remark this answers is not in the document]
tddough98: This hasn't been pruned, it's just been moved. The tracks alias has been moved to the TrackEdit class and the readable tracks is front-end logic as it's just converting a list of integers to a readable string

visualizer/src/service/raw/grayscaleMachine.js
```diff
@@ -35,7 +35,7 @@ const frameState = {
 };
 
 const channelState = {
-  initial: 'idle',
+  initial: 'loading',
```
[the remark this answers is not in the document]
tddough98: Most of the machines involved in loading image data having initial state `loading` (see imageMachine, labeledMachine. colorMachine, rawMachine, and this one grayscaleMachine). The app starts loading image data immediately by imageMachine with `entry: ['loadRaw', 'loadLabeled'],` so all the machines that anticipate events from loaded data need to start in the `loading` state so they don't miss these events by starting in `idle`

visualizer/src/service/labeled/labeledMachine.js
```diff
@@ -125,58 +109,28 @@ const createLabeledMachine = (projectId, numFeatures, numFrames) =>
...
-            );
-            return () => {
-              unbind('shift+f');
-              unbind('f');
-            };
-          },
         listenForHighlightHotkey: () => send => {
           bind('h', () => send('TOGGLE_HIGHLIGHT'));
           return () => unbind('h');
         },
```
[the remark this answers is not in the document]
tddough98: Created a new HighlightToggle switch and added it to the LabeledControls component. This switch now sets up the keybind

visualizer/src/service/imageMachine.js
```diff
@@ -210,33 +115,32 @@ const createImageMachine = ({ projectId }) =>
...
-        loadRaw: send(
-          ({ loadingFrame }) => ({ type: 'LOAD_FRAME', frame: loadingFrame }),
-          { to: ({ rawRef }) => rawRef }
-        ),
+        loadLabeled: send(({ loadingFrame }) => ({ type: 'LOAD_FRAME', frame: loadingFrame }), {
+          to: 'labeled',
+        }),
+        loadRaw: send(({ loadingFrame }) => ({ type: 'LOAD_FRAME', frame: loadingFrame }), {
+          to: 'raw',
+        }),
```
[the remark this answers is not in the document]
tddough98: that's right, see the tip right under this section in the XState docs [https://xstate.js.org/docs/guides/actors.html#sending-events-to-actors](https://xstate.js.org/docs/guides/actors.html#sending-events-to-actors)

deepcell_label/url_loaders_test.py
tddough98: I think the behavior of creating a project when no labels are present has changed from having the same same shape as the raw input vs. now only having a single feature. Previously, loading a raw image with two channels would generate a label array with two features, but now it's just one feature. Having the numbers of features change on with the number of channels felt like a weird dependency

willgraf (approved):
Nice work! 🚀

PR #260, opened 2021-09-08
Update named parameters for `replace_with_parent`.

willgraf:
## What
* Change the named arguments for `replace_with_parent` from the old `label_1`/`label_2` to just `daughter`.

## Why
* The backend was hitting 500 errors due to named argument errors.

### DIFF
--- a/deepcell_label/label.py
+++ b/deepcell_label/label.py
@@ -733,19 +733,38 @@ class TrackEdit(BaseEdit):
 
     def action_replace_with_parent(self, daughter):
         """
-        Replaces daughter with its parent after the division.
+        Replaces daughter with its parent after the division
+        and removes the division.
+        Future divisions with the daughter become part of the parent's lineage.
         """
-        daughter_division = self.tracks[daughter]
-        parent = daughter_division['parent']
-        parent_division = self.tracks[parent]
-        frame = parent_division['frame_div']
+        daughter_track = self.tracks[daughter]
+        parent = daughter_track['parent']
+        parent_track = self.tracks[parent]
+        frame = parent_track['frame_div']
 
+        # Remove division
+        for d in parent_track['daughters']:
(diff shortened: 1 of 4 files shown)


willgraf:
We should probably update the test suite to make sure actions sent with named args are valid.

epao111:
PR fixes the backend 500 error for replace with parent. The function now works but still leaves the daughter label associated with the parent. "remove from division" fixes this error but is not ideal for crowd sourcing.

tddough98:
I want to clarify a couple of cases for replace_with_parent, specifically 

- when the parent has multiple daughters
- and when the daughter divides as well.

Based on @epao111's comment, it seems like replace from parent should remove the division event entirely, even when other daughters are present. For example, when we have a division 1 -> 2 & 3, then replace 2 with 1 should remove 3 from the division too.

When the replace daughter also divides, I think that division should be associate|
For example, when we have divisions 1 -> 2 & 3 as well as 2 -> 4 & 5, then replace 2 with 1 should remove 3 from the division and associate 4 and 5 as daughters of 1. After we would have 1 -> 4 & 5 and 3 as a separate cell not part of any division.

willgraf:
> replace from parent should remove the division event entirely

Yes, I think that would be a good change. If the cell exists in the next frame then it cannot possibly divide. No reason to make the user remove the division event as an extra step.

I believe @epao111 also detailed some other bugs in the google doc.

epao111:
I think my comment was misinterpreted. When we have to combine to different labels for example one cell is labeled 1 in frames 1-5 and labeled 2 in frames 5-10, we would add 2 as the daughter of 1 and then perform the "replace with parent" action. Right now the action completes but label 2 stays the daughter of label 1 even thought it doesn't exist anymore. Additionally, the bar that indicates where the labels exist don't update.

PR #262, opened 2021-09-27
Update swap and replace hotkeys

tddough98:
Swap and replace can cause drastic changes to the labeling, in particular filling all the unlabeled areas when swapping or replacing the `0` label. Changing the hotkeys from S to Shift + S and R to Shift + R makes these actions harder to accidentally do.


willgraf (approved):
LGTM

PR #263, opened 2021-09-28
Add a form to the `Load` page to capture the image's dimension ordering.

willgraf:
## What
* Added `ImageAxesDropDown`, a new subcomponent that sets the `imageAxes` state of the `FileUpload` component.

## Why
* This allows the user to upload images of various dimension ordering.

---

Note: Because the page redirects once the file is uploaded, users must select the dimension ordering BEFORE they upload the file. If we want to do otherwise, we will need to change the `FileUpload` component to require a submit button.

Note 2: there is currently little validation done on the backend, so if the user submits the wrong dimension ordering, it can lead to quite a few backend requests.

### DIFF
--- a/deepcell_label/url_loaders.py
+++ b/deepcell_label/url_loaders.py
@@ -131,7 +131,7 @@ class FileLoader(Loader):
     def __init__(self, request):
         super(FileLoader, self).__init__()
         file_ = request.files.get('file')
-        self.input_axes = DCL_AXES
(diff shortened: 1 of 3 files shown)


willgraf:
I opted to leave the `URLLoader` using the `request.form` object, as the tests pass in a dictionary and I am not sure at the moment how to pass it a `flask.Request` object instead.

tddough98:
Thank you Will! This looks great. Adding a submit button to give time to update settings is a good idea, I'll work on that

tddough98 (approved):

PR #264, opened 2021-09-29
Track selected labels for undo/redo

tddough98:
I noticed when debugging Anolytic's issue this morning with not being able to make new labels that the selected labels were not changing when undoing and redoing. The segmentMachine used to manage the selected labels, so they were undone and redone through this machine, so it looks like we never added the selectMachine to the undo/redo history.

### DIFF
--- a/visualizer/src/service/selectMachine.js
+++ b/visualizer/src/service/selectMachine.js
@@ -1,4 +1,5 @@
 import { actions, assign, Machine, send, sendParent } from 'xstate';
+import { respond } from 'xstate/lib/actions';
 
 const { pure } = actions;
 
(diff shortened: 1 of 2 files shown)


visualizer/src/service/selectMachine.js
```diff
@@ -148,6 +151,12 @@ const selectMachine = Machine(
       ...selectShortcutActions,
       ...cycleActions,
       ...setActions,
+      save: respond(({ foreground, background }) => ({ type: 'RESTORE', foreground, background })),
+      restore: pure((_, { foreground, background }) => [
+        respond('RESTORED'),
+        send({ type: 'FOREGROUND', foreground }),
+        send({ type: 'BACKGROUND', background }),
+      ]),
```
willgraf, line 159: I expected to see a `SAVE` type in here somehwere?

tddough98: `SAVE` events are only sent by the `undoMachine` when an action happens to tell a tracking machine to create a snapshot of itself that it restore in case the action is undone/redone. When sending `SAVE`, the `undoMachine` expects tracked machines to respond with a `RESTORE` event that the `undoMachine` records and sends back to the tracked machine when undoing or redoing the action

willgraf: thank you for the clarification!

willgraf (approved):
LGTM

PR #265, opened 2021-10-01
Fix tool keybinds

tddough98:
When using tool hotkeys to switch out the brush and threshold currently causes UI crashes as the tool changes before the `BrushCanvas` or `ThresholdCanvas` is removed, causing these canvases to access a brush machine or threshold machine that no longer exists. 

Previously, we spawned one tool machine at a time and deleted the previous tool machine when switch between tools. With this PR, we spawn all of the tools and keep them in the `segmentMachine` and switch between these pre-existing tool. Instead of a `useTool` hook that provides the current tool, we have specific `useBrush` and `useThreshold` hooks that provide the persistent brush or threshold actors so these components don't crash even if the component is out of sync with the current tool.

This PR also includes a couple of fixes and improvements that I noticed while making these changes.

 - refactoring switching between tools in `segmentMachine`
 - fixing the flood tool to flood immediately instead of first selecting the background and then flooding
 - fixing double clicks with the select tool to deselect the foreground because is was sending the wrong event
 - changing machines to start with `null` in their context and sending events to initialize them to

### DIFF
--- a/visualizer/src/service/tools/segment/floodMachine.js
+++ b/visualizer/src/service/tools/segment/floodMachine.js
@@ -1,40 +1,39 @@
 import { Machine, sendParent } from 'xstate';
 import { toolActions, toolGuards } from './toolUtils';
 
-const createFloodMachine = () =>
-  Machine(
-    {
-      context: {
-        x: null,
-        y: null,
-        label: null,
-        foreground: null,
-        background: null,
-      },
-      on: {
-        COORDINATES: { actions: 'setCoordinates' },
-        HOVERING: { actions: 'setHovering' },
-        FOREGROUND: { actions: 'setForeground' },
-        BACKGROUND: { actions: 'setBackground' },
-        mouseup: { actions: ['flood', 'selectBackground'] },
-      },
+const floodMachine = Machine(
+  {
+    context: {
+      x: null,
+      y: null,
+      label: null,
+      foreground: null,
+      background: null,
     },
-    {
-      guards: toolGuards,
-      actions: {
-        ...toolActions,
-        selectBackground: sendParent('SELECT_BACKGROUND'),
-        flood: sendParent(({ foreground, x, y }, event) => ({
-          type: 'EDIT',
-          action: 'flood',
-          args: {
-            label: foreground,
-            x_location: x,
-            y_location: y,
-          },
-        })),
-      },
-    }
-  );
+    on: {
+      COORDINATES: { actions: 'setCoordinates' },
+      HOVERING: { actions: 'setHovering' },
+      FOREGROUND: { actions: 'setForeground' },
+      BACKGROUND: { actions: 'setBackground' },
+      mouseup: { actions: ['flood', 'selectBackground'] },
+    },
+  },
+  {
+    guards: toolGuards,
+    actions: {
+      ...toolActions,
+      selectBackground: sendParent('SELECT_BACKGROUND'),
+      flood: sendParent(({ foreground, x, y }, event) => ({
+        type: 'EDIT',
+        action: 'flood',
+        args: {
+          label: foreground,
+          x_location: x,
+          y_location: y,
+        },
+      })),
+    },
+  }
+);
 
-export default createFloodMachine;
+export default floodMachine;
--- a/visualizer/src/service/tools/segment/floodMachine.js
+++ b/visualizer/src/service/tools/segment/floodMachine.js
@@ -1,22 +1,22 @@
 import { Machine, sendParent } from 'xstate';
 import { toolActions, toolGuards } from './toolUtils';
 
-const createFloodMachine = ({ x, y, label, foreground, background }) =>
+const createFloodMachine = () =>
   Machine(
     {
       context: {
-        x,
-        y,
-        label,
-        foreground,
-        background,
+        x: null,
+        y: null,
+        label: null,
+        foreground: null,
+        background: null,
       },
       on: {
         COORDINATES: { actions: 'setCoordinates' },
-        LABEL: { actions: 'setLabel' },
+        HOVERING: { actions: 'setHovering' },
         FOREGROUND: { actions: 'setForeground' },
         BACKGROUND: { actions: 'setBackground' },
-        mouseup: [{ cond: 'onBackground', actions: 'flood' }, { actions: 'selectBackground' }],
+        mouseup: { actions: ['flood', 'selectBackground'] },
       },
     },
     {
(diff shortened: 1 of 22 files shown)


visualizer/src/service/tools/segment/selectMachine.js
```diff
@@ -1,37 +1,36 @@
...
+      FOREGROUND: { actions: 'setForeground' },
+      BACKGROUND: { actions: 'setBackground' },
+      HOVERING: { actions: 'setHovering' },
+      mouseup: [
+        {
+          cond: 'doubleClick',
+          actions: ['resetForeground', 'selectBackground'],
+        },
+        { cond: 'onForeground', actions: 'selectBackground' },
+        { actions: [(c, e) => console.log(c), 'selectForeground'] },
```
willgraf, line 21: Let's ditch the `console.log`

visualizer/src/service/tools/segment/thresholdMachine.js
```diff
@@ -1,52 +1,52 @@
...
-            y2: y,
-            // frame: context.frame,
-            label: foreground,
-          },
-        })),
-      },
-    }
-  );
+    on: {
+      EXIT: '.idle',
```
willgraf, line 26: Why does the Brush and Threshold machine need the `EXIT` event, but not the Flood, Select, Trim, or Watershed?

visualizer/src/service/tools/segmentMachine.js
```diff
@@ -198,16 +143,36 @@ const segmentMachine = Machine(
...
       ...toolGuards,
-      usingColorTool: ({ tool }) => colorTools.includes(tool),
+      usingColorTool: ({ tool }) => ['brush', 'select', 'trim', 'flood'].includes(tool),
+      isColorTool: (_, { tool }) => ['brush', 'select', 'trim', 'flood'].includes(tool),
+      usingGrayscaleTool: ({ tool }) =>
+        ['brush', 'select', 'trim', 'flood', 'threshold', 'watershed'].includes(tool),
+      isGrayscaleTool: (_, { tool }) =>
+        ['brush', 'select', 'trim', 'flood', 'threshold', 'watershed'].includes(tool),
+      isNoPanTool: (_, { tool }) => ['brush', 'threshold'].includes(tool),
+      isPanTool: (_, { tool }) => ['select', 'trim', 'flood', 'watershed'].includes(tool),
```
willgraf, line 153: Not wild about all these lists, but not a blocker for this PR. Would be nice to classify these in a more clear way.

visualizer/src/service/tools/segmentMachine.js
```diff
@@ -1,37 +1,39 @@
...
-  threshold: createThresholdMachine,
-  trim: createTrimMachine,
-  flood: createFloodMachine,
-  watershed: createWatershedMachine,
-};
-
-const createToolMachine = context => {
-  const { tool } = context;
-  return spawn(createToolMachineLookup[tool](context), 'tool');
+const panState = {
```
willgraf, line 12: Seems like a lot of changes in this file - can you briefly summarize?

willgraf (changes requested):
I left a few clarification comments, and one actual change request (remove console.log). A few other notes:

- I love the tool/machine refactoring! The state machine initialization makes a lot of sense, and breaking out brush and others into their own hooks is great 💯 
- `label` relabeled to `hovering` makes me a bit nervous that we missed one somewhere

visualizer/src/service/tools/segment/toolUtils.js
```diff
@@ -3,16 +3,16 @@ import { assign } from 'xstate';
...
 export const toolActions = {
   setCoordinates: assign((_, { x, y }) => ({ x, y })),
-  setLabel: assign((_, { label }) => ({ label })),
-  setForeground: assign({ foreground: (_, { foreground }) => foreground }),
-  setBackground: assign({ background: (_, { background }) => background }),
-  setSelected: assign({ selected: (_, { selected }) => selected }),
+  setHovering: assign((_, { hovering }) => ({ hovering })),
+  setForeground: assign((_, { foreground }) => ({ foreground })),
+  setBackground: assign((_, { background }) => ({ background })),
+  setSelected: assign((_, { selected }) => ({ selected })),
```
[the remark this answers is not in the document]
tddough98: This was to make each line consistent. There are two ways assign actions can be written
```assign({ foreground: (_, ({ foreground }) => foreground })```
or 
```assign((_, { foreground }) => ({ foreground }))```
Here I changed all of them to the later because it was mixed before

File: visualizer/src/service/tools/segment/thresholdMachine.js (hunk 2 above)
Comment (tddough98) [reply visualizer/src/service/tools/segment/thresholdMachine.js:26]: The brush and threshold machines have internal finite state that we need to reset when changing tools midway through using them. The other machines also receive an EXIT event when switching tools, but they don't need to use them.

For example, if we start painting a brush stroke, then switch tools, if we did not send the EXIT event, then when we start using the brush tool again, we would still be painting that same brush stroke

Comment (tddough98) [reply visualizer/src/service/tools/segment/thresholdMachine.js:26]: Actually now that I say that, watershed needs to use EXIT because it has two steps

File: visualizer/src/service/tools/segmentMachine.js (hunk 3 above)
Comment (tddough98) [reply visualizer/src/service/tools/segmentMachine.js:153]: I'll put them into const lists at the top of the file and reference them here

File: visualizer/src/service/tools/segment/floodMachine.js [hunk 6]
```diff
@@ -1,40 +1,39 @@
...
-        })),
-      },
-    }
-  );
+    on: {
+      COORDINATES: { actions: 'setCoordinates' },
+      HOVERING: { actions: 'setHovering' },
+      FOREGROUND: { actions: 'setForeground' },
+      BACKGROUND: { actions: 'setBackground' },
+      mouseup: { actions: ['flood', 'selectBackground'] },
```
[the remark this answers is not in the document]
tddough98: yes this was one of the changes I noted in the top level PR description.

> fixing the flood tool to flood immediately instead of first selecting the background and then flooding

visualizer/src/service/tools/segment/selectMachine.js
tddough98: whoops left behind from debugging

visualizer/src/service/tools/segmentMachine.js
tddough98: - child tool machines no longer need the parent context to be spawned, so the imports are renamed from `createXMachine to `xMachine`
- spawning child tools is consolidated into a single entry action `spawnTools` instead of happening every time we switch tools
- new `panState` to check if we should send a `SET_PAN_ON_DRAG` event to the canvas machine
- new `syncState` that groups all the state update events to forward to child tool actors
- combine `USE_X` events into `SET_TOOL` event
- initialized the context with `null` and then syncing with events from other actors instead of hard coding multiple machines to start with the same context

willgraf (approved):
LGTM.

Issue #29, opened 2019-07-31
Readme should have more information for users


geneva-miller:
Effects of different commands should be explained, and list of commands should be expanded as functionality expands. Will also need significant update for other Caliban use modes (eg, zstack editing).

tddough98:
Closed by #248 and #266

PR #266, opened 2021-10-05
Update documentation

tddough98:
This updates the README for DeepCell Label to be up to date with the current master branch. As this only changes the markdown README at the front of the repo, I'd love to merge it in before tomorrow morning's demo. I'm also branching off this to update the in app instructions, but as that changes the in app code and React components, I'll keep that separate and merge it after more review.


willgraf:
I hope this PR can also fix #29 (which may be able to be closed already).

tddough98:
Failing node tests are dealt with in #269. Once that is in master and then merged into this branch, I expect the checks to pass

willgraf (commented):
Is there a way to shrink the size of some of the PNGs? The **Editing Labels** section in particular is distractingly large IMO.

willgraf (commented):
Just a few comments:

- README looks great!
- Do we still need the PNGs?
- Why did you need to remove all the key-unbinds?
- I wanted to spin up locally to look at the instructions, but I keep crashing when I click the dropdown:
Screen Shot 2021-11-01 at 10 25 12 AM

Any suggestions?

tddough98:
We do not need the PNGs, removed them now

I removed the unbinds because the instructions were unbinding them when closing the tab, so the keybinds would only work when the page first loads or when the the instructions are open.

Looks like the crash is happening because the instructions loads components that expect state machines to exist that are only created once the project is loaded. My usual strategy here is to check if the stateMachine exists with `{stateMachine && <ComponentThatUsesStateMachine />}`. I'll add these in as needed here. But you'll also need to look at an actual project to see the components in the instructions. Now that the landing page is merged that should be easy as you can make an example project from there

willgraf (approved):
LGTM!

I had a thought while looking at the instructions: maybe a accordion/dropdown would work for each header of the instructions? some of the instructions are quite long, the accordion could help minimize the size? (at least initially). 

See the [deepcell.org about page](https://deepcell.org/about) for the accordion effect I'm thinking of

PR #268, opened 2021-10-20
Add landing page

tddough98:
Here we add a landing page to DeepCell Label with an introduction to the app, a dropzone to upload TIFFs, NPZs and PNGs, and dropdown menu with two example files. The landing page is a more user-friendly entry point to using Label, giving users another option than the Label API to create projects.

Here's an overview of the changes:

- created a base Loader class and separate URLLoader and FileLoader implements
- the app now has two pages: `/` for landing page and `/project` for interacting with a project
   - this is a breaking change for Anolytics jobs as the we'll have to add `/project` to our job URLs
- the project page can show a download button instead of a submit button by adding a `download=true` URL parameter

There are also some smaller bug fixes and improvements included:

- loading PNGs is more flexible and robust, now handling palette-based and grayscale PNGs
- the watershed state machine no longer exits its clicked state immediately due to the state selecting the foreground label
- controls for white channels are now shown in black instead so they don't blend into the background


visualizer/src/service/apiMachine.js
```diff
@@ -85,10 +108,27 @@ const createApiMachine = ({ projectId, bucket }) =>
             },
           },
         },
```
tddough98, line 110: Another machine viz. The change is the new downloading state

Screen Shot 2021-10-20 at 10 34 57 AM

deepcell_label/url_loaders_test.py
```diff
@@ -10,7 +10,7 @@
 from skimage.external.tifffile import TiffWriter
 import zipfile
 
-from deepcell_label import url_loaders
+from deepcell_label.url_loaders import URLLoader, FileLoader
```
willgraf, line 13: Does FileLoader get tested?

visualizer/src/App.js
```diff
@@ -13,11 +14,14 @@ function App() {
...
       <Switch>
-        <Route path='/'>
+        <Route path='/project'>
           <ProjectContext project={service}>
             <Label />
           </ProjectContext>
         </Route>
+        <Route path='/'>
+          <Load />
+        </Route>
```
willgraf, line 24: Minor preference: I'd like it if the order was switched for clarity. (root path first)

willgraf (commented):
Content looks great, but would it be possible to add additional testing?

It would be nice to test some basics on the new react components, but I definitelyly think all the backend changes should be tested (mainly talking about `FileLoader`)

visualizer/src/App.js
tddough98: I think that changes the behavior

From the [React Router documentation](https://reactrouter.com/web/guides/quick-start)

> A <Switch> looks through its children <Route>s and renders the first one that matches the current URL.

so `/` would match everything. The examples on that page put `/` last

deepcell_label/blueprints_test.py
```diff
@@ -4,8 +4,9 @@
 
 import pytest
 import numpy as np
+from skimage.external.tifffile import TiffWriter
```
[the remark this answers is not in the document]
tddough98: Changed to tifffile. I had to pin the tifffile version to 2020.9.3 because that's the last version to support Python 3.6, but I'll update that in another PR to fully drop support for Python 3.6

willgraf (approved):
LGTM! 🚀

PR #275, opened 2021-11-03
Reduce Docker image with slim base image and multi-stage builds & small code quality improvements.

willgraf:
## What
* Remove unused imports from various backend files.
* Fix `exporters_test.py`, use `with app.app_context()` in the tests directly, remove test fixtures.
* Add `.dockerignore`
* Update `Dockerfile` to use `python:3.7-slim-bullseye` and multi-stage builds for a smaller final image.
* Install and use `gunicorn` as the production server inside the Dockerfile
* Update README to format the code blocks as bash (resolves markdown warning).
* Upgrade Python dependencies to latest patch releases and allow future patches to be used.

## Why
* Incremental improvements to code quality.
* Build a better Dockerfile to serve as the backend. This should make migration from EB to ECS (or others) easier.


willgraf:
I also want to note that I think we are pinned to using Python 3.7 due to the `scikit-image` version restriction.

tddough98:
As the dockerimage is for the backend only, should we ignore `/visualizer`?

.dockerignore
```diff
@@ -0,0 +1,143 @@
...
+.DS_Store
+.vscode
+.gitignore
+.dockerignore
+
+/credentials
+/cache
+/store
+
+/node_modules
```
tddough98, line 12: As the dockerimage is for the backend only, should we ignore `/visualizer` in .dockerignore?

tddough98:
Should we add instructions to the README on how to build and use the docker image? We could wait on that until we're using a container in production

.dockerignore
willgraf: Good call, its been added.

willgraf:
README changes are a good call. I broke up the README Flask portion - take a look and LMK what you think.

PR #276, opened 2021-11-04
Upgrade to react-router v6

tddough98:
React Router released a [new major version 6](https://remix.run/blog/react-router-v6) which has more flexible hooks and is more composable. It replaces the <Switch> component with <Routes> which avoids bugs due to route ordering and lets us place the `/` route before the `/project` route, which @willgraf requested in a recent review.

### DIFF
--- a/visualizer/package.json
+++ b/visualizer/package.json
@@ -36,7 +36,7 @@
     "react-dom": "^17.0.1",
     "react-dropzone": "^11.4.0",
     "react-icons": "^4.2.0",
-    "react-router-dom": "6",
+    "react-router-dom": "^6.0.0",
     "react-scripts": "4.0.2",
     "typescript": "^4.2.4",
     "web-vitals": "^1.0.1",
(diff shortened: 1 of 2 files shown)


visualizer/package.json
```diff
@@ -36,7 +36,7 @@
     "react-dom": "^17.0.1",
     "react-dropzone": "^11.4.0",
     "react-icons": "^4.2.0",
-    "react-router-dom": "^5.2.0",
+    "react-router-dom": "6",
```
willgraf, line 39:
```suggestion
    "react-router-dom": "^6.0.0",
```

I think it's better to make sure we are able to build with patch releases.

Also, is there any dependency on the react version itself and react-router v6?

willgraf (changes requested):
Looks like the test failures are a 3rd party coveralls issue. Maybe we can pin the workflow coveralls version? I haven't looked into it but seems like there was probably a breaking change in a new release.

visualizer/package.json
tddough98: It requires React >=16.8 which introduced that and our React version dependency is already higher. I think react-router-dom is react-router for the browser and there's also react-router-native for react native. The [documentation says to install react-router-dom](https://reactrouter.com/docs/en/v6/getting-started/overview)

willgraf (approved):

PR #277, opened 2021-11-04
Put a version ceiling on coveralls, 3.3.0 has a bug preventing builds.

willgraf:
## What
* Pin the `coveralls` version to be < 3.3.0.

## Why
* `coveralls` released [version 3.3.0](https://pypi.org/project/coveralls/) yesterday which is causing [builds to break](https://github.com/vanvalenlab/deepcell-label/runs/4111480977?check_suite_focus=true). This change prevents the new version from being installed.


tddough98 (approved):
Awesome, thanks for working out the fix

Issue #280, opened 2021-11-26
Label number goes to NaN


GMaciag:
**Describe the bug**
Cannot add more than one label when annotating your own image in label.deepcell.org 
The label number after "1" goes to NaN. The Palette keeps showing NaN, you cannot switch between annotations, you cannot add new ones. The NaN annotation ends up hidden "under" the image (increase opacity to see).

**To Reproduce**
1. Go to [label.deepcell.org ](url).
2. Load your own image.
3. You can annotate the first cell, it will be labelled "1" as expected. 
3. Press 'n' (or select the plus sign in the Foreground/Background Palette).
4. The cell number is now NaN.

**Expected behavior**
You should be able to create multiple cell annotations.

**Screenshots**

<img width="1665" alt="Screenshot 2021-11-26 at 13 58 17" src="https://user-images.githubusercontent.com/31366577/143585051-30a3da17-7c9c-4e64-bb22-5bcde556f1d2.png">

**Additional context**
The issue does not occur when working on an example dataset included in the website.

tddough98:
[label.deepcell.org](label.deepcell.org) has now been fixed so you should be able to add new labels to any file now. Thank you for raising this issue and please let us know if you have any other issues or questions come up!

PR #281, opened 2021-11-29
Combine ZStackEdit and TrackEdit classes to ensure consistent `cell_info` contents

tddough98:
Since we've added tracking to the current build of DeepCell Label, we've been creating tracking metadata for all projects so that any project could use the tracking code and UI. However, the edit route still distinguished between tracking and zstack projects using the extension of the file used to create the project. Only files ending with .trk or .trks are edited with TrackEdit, and all other files are edited with ZStackEdit. As each class implements its own `add_cell_info` and `del_cell_info`, editing with the ZStackEdit would not maintain the tracking metadata as expected.

Issue #280 revealed this issue, as a project with no initial labels would have an empty `cell_info` dict, and once we start adding labels using the ZStackEdit class, the `cell_info` would not have the expected fields. The route that serves `cell_info` expects these fields to be present, and serves an error instead and causing NaN to be appear the selected label.

Here the TrackEdit and ZStackEdit classes are unified into a single Edit class so that all projects are both initialized and updated with the same consistent set of field in its `cell_info` dict. I also removed actions that are not currently used.


visualizer/src/service/apiMachine.js
```diff
@@ -32,19 +33,26 @@ function redo(context, event) {
...
+  const format = new URLSearchParams(window.location.search).get('track') ? 'trk' : 'npz';
+  const promise = axios.get('/api/download', {
+    params: { id: projectId, format: format },
+    responseType: 'blob',
+  });
+  promise.then(response => console.log(response));
   const filename = promise.then(response => {
     const regex = /filename=(.*)$/;
-    const header = response.headers.get('content-disposition');
+    const header = response.headers['content-disposition'];
```
willgraf, line 55: Why the switch from `fetch` to `axios`?

deepcell_label/utils.py
```diff
@@ -0,0 +1,44 @@
...
+            parent = info['parent']
+            info['frame_div_parent'] = new_info[parent]['frame_div']
+        else:
+            info['frame_div_parent'] = None
+    return new_info
+
+
+def snakecase_to_camelcase(name):
+    snake_pattern = re.compile(r'_([a-z])')
+    return snake_pattern.sub(lambda x: x.group(1).upper(), name)
```
willgraf, line 27: Might as well throw in a test for this too.

willgraf (changes requested):
Looks awesome, a great improvement. In addition to a couple of small questions, it looks like there is some `etag` test failure going on.

visualizer/src/service/apiMachine.js
tddough98: Switched back to fetch

willgraf (approved):
💯

PR #282, opened 2021-11-29
Add tracking and QC available on label.deepcell.org to main branch

tddough98:
The deployed app on label.deepcell.org is ahead of the main branch as it contains tracking and quality control features.

Code to support tracking includes

- actions in label.py like `add_daughter`, `remove_daughter`, and `replace_with_parent`, 
- tests for these actions
- Tracking UI components in `visualizer/src/Controls/Tracking`
- `trackMachine.js` to send events to edit the lineage
- `toolMachine.js` to switch between `trackMachine` to edit the lineage and `segmentMachine` to edit the segmentation

Code to support quality control includes

- QC components in `visualizer/src/Controls/QualityControl`
- `qualityControlMachine` to make judgments on projects

Other small improvements include

- adding ETags to routes to avoid unnecessarily resending data, 
- cleaning label arguments for actions to be valid integers between 0 and a new label


visualizer/src/Controls/Tracking/Alerts/AlertGroup.js
```diff
@@ -0,0 +1,78 @@
...
+    headerAlert: {
+      boxSizing: 'border-box',
+      maxWidth: '300px',
+      marginTop: theme.spacing(2),
+    },
+    alert: {
+      position: 'relative',
+      boxSizing: 'border-box',
+      maxWidth: '280px',
+      left: '20px',
```
willgraf, line 30: Minor: I'm not a big fan of raw pixels, it would be great to use relative units if possible.

visualizer/src/Controls/Tracking/Division.js
```diff
@@ -0,0 +1,251 @@
...
+      </Popper>
+    </Box>
+  );
+}
+
+function DaughterMenu({ parent, daughter }) {
+  const tracking = useTracking();
+  const { send } = tracking;
+
+  const [open, toggle] = useReducer((v) => !v, false);
```
willgraf, line 165: Minor: IDK if this naming is standard, but it might be useful to change the `toggle` to `toggleOpen` (or whatever we are toggling). It wasn't clear to me at first.

visualizer/src/Controls/Tracking/LabelTimeline.js
```diff
@@ -0,0 +1,65 @@
...
+
+function FrameBox({ frame, numFrames, color }) {
+  const image = useImage();
+
+  const boxStyle = {
+    position: 'absolute',
+    backgroundColor: color,
+    height: '10px',
+    width: `${(1 / numFrames) * 100}%`,
+    left: `${(frame / numFrames) * 100}%`,
```
willgraf, line 22: nice

visualizer/src/Controls/Tracking/LabelTimeline.js
```diff
@@ -0,0 +1,65 @@
...
+          numFrames={numFrames}
+          color={frames.includes(frame) ? color : '#FFFFFF'}
+          key={frame}
+        />
+      ))}
+    </Box>
+  );
+}
+
+export default React.memo(LabelTimeline);
```
willgraf, line 65: Can you comment on why this is a `memo`?

visualizer/src/service/service.js
```diff
@@ -1,13 +1,49 @@
...
+function createProject() {
+  const location = window.location;
+  const search = new URLSearchParams(location.search);
+  const projectId = search.get('projectId');
+  if (!projectId || projectId.split(',').length > 1) {
+    return;
+  }
+  const bucket = search.has('bucket') ? search.get('bucket') : 'caliban-output';
+  const machine = createProjectMachine(projectId, bucket);
+  const service = interpret(machine); // , { devTools: true });
```
willgraf, line 26: Minor: Does React have a debug vs production mode? Maybe we can leave the `devtools` in and set the value to be "isDebugMode"?

visualizer/src/Controls/Tracking/Alerts/AlertGroup.js
```diff
@@ -0,0 +1,78 @@
+import { IconButton } from '@material-ui/core';
+import Collapse from '@material-ui/core/Collapse';
+import { makeStyles } from '@material-ui/core/styles';
+import ExpandMoreIcon from '@material-ui/icons/ExpandMore';
+import Alert from '@material-ui/lab/Alert';
+import clsx from 'clsx';
```
willgraf, line 6: What is `clsx`?

willgraf (approved):
I love it! 💯 very readable code, well-named classes so it is very clear. Easy approval.

I do have some questions/comments but they are all very minor and are non-blocking.

visualizer/src/service/service.js
tddough98: We could use `process.env.NODE_ENV` to distinguish between dev and prod

> There is also a built-in environment variable called NODE_ENV. You can read it from process.env.NODE_ENV. When you run npm start, it is always equal to 'development', when you run npm test it is always equal to 'test', and when you run npm run build to make a production bundle, it is always equal to 'production'. You cannot override NODE_ENV manually. 

(from [https://create-react-app.dev/docs/adding-custom-environment-variables/](https://create-react-app.dev/docs/adding-custom-environment-variables/))

It looks like we can't add a debug mode on top of these, and I often do not want the devtools enabled as they can slow things down significantly. I uncomment the devtools occasionally to visual inspect machines, but I think leaving it commented out is correct so that it's not unintentionally enabled by new developers.

visualizer/src/Controls/Tracking/Alerts/AlertGroup.js
tddough98: It's a library for manipulating classNames: [https://www.npmjs.com/package/clsx](https://www.npmjs.com/package/clsx)

We're using it to add formatting when opening the alert boxes

It's installed as part of our dependencies, but I'm realizing that it's not explicitly in our package.json, so I'll add that in

visualizer/src/Controls/Tracking/LabelTimeline.js
tddough98: The divisions UI tends to rerender too often. For example, when moving the mouse over a different label on the canvas, the React dev tools show these components rerendering because some of their hooks changed, but no data or state relevant to the division changed, so ideally the components should not rerender. While developing, I was hoping that memo could help out with this issue. After revisiting it, it seems that the component still rerenders even though it is memoized...

The React component profiler shows that the rerenders take about 10 milliseconds, which should not be noticeable as it's less than the 16 ms between frames, but it could contribute to jank if are other components rerendering at the same time. I'll post an issue to keep track of this so we can fix this in a future release.

visualizer/src/Controls/Tracking/Alerts/AlertGroup.js
tddough98: I switched pixels to rem and I moved these reused styles into AlertGroup.js and imported them in other alerts like OneDaughterAlert.js.

visualizer/src/Controls/Tracking/LabelTimeline.js
tddough98: Added issues #293 to track it outside of this PR

visualizer/src/Controls/Tracking/Alerts/AlertGroup.js
willgraf: Importing the styles really helped clean it up!

willgraf (approved):
LGTM!

PR #283, opened 2021-11-30
Add route to add channels to an existing project

tddough98:
Here we add an API route to add additional channels to a project. This supports displaying predictions from Impartial which creates probability maps from scribbles created in DeepCell Label, so these maps need to be added later in the project lifecycle. By showing the Impartial results in the same project, a user can compare their scribbles to Impartial predictions and update the scribbles to repeat the process.

Create an npz with the new channel with shape (frames, height, width, 1) and post the channel to the `/api/raw/{projectId}` route with the npz as an attached file.
For example, with curl this looks like
```
curl --location --request POST 'label.deepcell.org/api/raw/EXAMPLE-ID' --form 'file=@"newChannel.npz"'
```

Areas for improvement could include loading channels from more filetypes, specifying a flexible dimension order, and adding multiple channels with one request.


deepcell_label/models.py
```diff
@@ -454,6 +456,19 @@ def get_labeled_array(self, feature, frame):
...
 
+    def add_channel(self, channel):
+        """
+        Adds a channel to an existing project.
+        """
+        new_raw_frames = []
+        for i, raw_frame in enumerate(self.raw_frames):
+            old_array = raw_frame.frame
+            new_array = np.concatenate((old_array, channel[i]), axis=-1)
+            new_raw_frames += [RawFrame(i, new_array)]
```
willgraf, line 467: I think the `.append` method is much faster/lighter. The concatenate operation ends up creating additional lists.

deepcell_label/blueprints.py
```diff
@@ -64,6 +65,24 @@ def raw(token, channel, frame):
...
+    # Load channel from first array in attached file
+    npz = np.load(request.files.get('file'))
+    channel = npz[npz.files[0]]
+    # Check channel is the right shape
+    expected_shape = (project.num_frames, project.width, project.height, 1)
+    if channel.shape != expected_shape:
+        raise ValueError(f'New channels for project {token} must have shape {expected_shape}')
+    # Add channel to project
+    project.add_channel(channel)
+    return project.make_first_payload()
```
willgraf, line 83: Just curious - why `make_first_payload`? We've already sent a first payload before this, right?

deepcell_label/blueprints.py
```diff
@@ -64,6 +65,24 @@ def raw(token, channel, frame):
...
 
+@bp.route('/api/raw/<token>', methods=['POST'])
+def add_raw(token):
+    """Add new channel to the project."""
+    project = Project.get(token)
+    if not project:
+        return abort(404, description=f'project {token} not found')
+    # Load channel from first array in attached file
+    npz = np.load(request.files.get('file'))
+    channel = npz[npz.files[0]]
```
willgraf, line 76: Consider wrapping this in a try/except. What if the request doesn't have the `file`?

willgraf (changes requested):
Looks pretty good, just a few small comments.

deepcell_label/blueprints.py
tddough98: We send `numChannels` in the first payload and that changes when we add a channel. I was thinking this could be useful in the future to automatically load the new channel if calling the add channel route from the client. I'll make this more compact to only contain the `numChannels` so it's clear that it is different and useful.

deepcell_label/blueprints.py
tddough98: I added blocks to catch `BadRequestKeyError` when the attached form does not have a file field and to catch `TypeError` for when `np.load` can't read the attached file.

willgraf (approved):
Looks good!

Issue #284, opened 2021-12-03
Unable to run flask 


StephanDrothler:
Seemingly, I successfully installed the requirements.txt in my virtual environment (I changed 
pillow>=7.1.0,<8.3.2 to pillow==8.3.2, otherwise a had an unresolvable version conflict).
Subsequently, I executed "flask run" and the following error message popped up:

"*Environment: production
WARNING: This is a development server. Do not use it in a production deployment.
Use a production WSGI server instead.
*Debug mode: off
Usage: flask run [OPTIONS]

Error: Could not locate a Flask application. You did not provide the "FLASK_APP" environment variable, and a "wsgi.py" pr "app.py" module was not found in the current directory."

How do I set the correct environment variable and provide the right modules?

tddough98:
We've an included an example `.env` file in the repo named `.env.example` that has the FLASK_APP environment variable prepopulated. If you remove the `.example` suffix from that file so the name is just `.env`, then those environment variables will be used by the app. We use the `.example` suffix to avoid accidentally commiting the real `.env` to the project, as it often contains sensitive environment variables like access keys or database passwords.

You can also set the environment variable without using a `.env` file by running the command
```bash
export FLASK_APP=application
```
Here's a [blog post](https://opensource.com/article/19/8/what-are-environment-variables) about environment variables with some more context and tips.

I'll add to the README instructions to make an `.env` file from `.env.example` and then close this issue. Thank you for posting this, we appreciate everything that can make the project more accessible and useful! Let me know if you have any other questions or issues and I'll be glad to help out.

PR #285, opened 2021-12-03
Add instructions on .env to README

tddough98:
Issue #284 had issues with setting environment variables. Adding some instructions to the readme on how to use the example `.env.example` will point users to the resources to configure their envvars correctly.

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,6 +27,10 @@ cd deepcell-label
 
 The backend is a Flask HTTP API that will store and update project data.
 
+#### Set up environment
+
+Flask expects some environment variables to be set, like FLASK_APP. Set these with `export FLASK_APP=application` in your terminal, or set them in the `.env` file in the root of the repository. There is an example `.env.example` with FLASK_APP already set. Make a copy as remove `.example` so the file is named `.env`.
+
 #### Run Python locally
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -27,6 +27,10 @@ cd deepcell-label
 
 The backend is a Flask HTTP API that will store and update project data.
 
+#### Set up environment
+
+Flask expects some environment variables to be set, like FLASK_APP. Set these with `export FLASK_APP=application` in your terminal, or set them in the `.env` file in the root of the repository. There is an example `.env.example` with FLASK_APP already set. Make a copy as remove `.example` so the file is named `.env`.
```
willgraf, line 32:
```suggestion
Flask expects some environment variables to be set, like FLASK_APP. Set these with `export FLASK_APP=application` in your terminal, or set them in the `.env` file in the root of the repository. There is an example `.env.example` with FLASK_APP already set. Make a copy and remove `.example` so the file is named `.env`.
```

willgraf: also, consider adding a brief explanation of why `application` is the correct value. (Or link to the flask docs that talks about the FLASK_APP var)

willgraf (changes requested):

willgraf (approved):

PR #286, opened 2021-12-06
Log frontend and backend git commits and branch names in browser

tddough98:
With parallel deployments of DeepCell Label, the specific versions deployed are hard to keep track of. As updates may have features or bug fixes relevant to a deployment, it's important to know whether these code changes are online.

Here, we add environment variables with the commit and branch name at build time for the frontend assets. The backend has a new route to serve its commit and branch name.

The log messages in the browser debug console look like:
Screen Shot 2021-12-06 at 10 52 49 AM
Screen Shot 2021-12-06 at 10 53 02 AM


deepcell_label/blueprints.py
```diff
@@ -33,6 +34,33 @@ def health():
...
+
+
+def get_git_branch() -> str:
+    return (
+        subprocess
+        .check_output(['git', 'rev-parse', '--abbrev-ref', 'HEAD'])
+        .decode('ascii')
+        .strip()
+    )
+
```
willgraf, line 53: What does this do if `git` is not installed or the `.git` folder can't be accessed?

Is there a way we can set the commit/branch in `config.py` and return them here?

PR #287, opened 2021-12-07
Migrate Elastic Beanstalk to Python 3.8 running on Amazon Linux 2

tddough98:
Our Elastic Beanstalk instances have been running on a deprecated platform Python 3.6 running on 64bit Amazon Linux. Here, we migrate to a supported platform Python 3.8 running on Amazon Linux 2, updating the EB configuration to work with the new platform.

This PR also drops support for Python 3.6, which reaches its [end of life and will stop receiving security updates in 2 weeks on December 23 2021](https://endoflife.date/python). I bumped all our dependencies and removed any references to Python 3.6. I did not add Python 3.9 and 3.10 to our testing workflow as there are only Python 3.7 and 3.8 platforms available on Elastic Beanstalk at the moment.


.ebextensions/pip.config
```diff
@@ -1,9 +1,9 @@
 container_commands:
   01_upgrade_pip:
-    command: "python -m pip install --upgrade pip"
+    command: "source /var/app/venv/*/bin/activate && python -m pip install --upgrade pip"
     leader_only: true
-  02_upgradde_setuptools:
-    command: "python -m pip install --upgrade setuptools"
+  02_upgrade_setuptools:
+    command: "source /var/app/venv/*/bin/activate && python -m pip install --upgrade setuptools"
```
willgraf, line 6: Is this still recommended?

.github/workflows/tests.yaml
```diff
@@ -60,7 +60,7 @@ jobs:
 
     strategy:
       matrix:
-        python-version: [3.6, 3.7, 3.8]
+        python-version: [3.7, 3.8]
```
willgraf, line 63: May as well add 3.9 as I assume we will one day need to migrate again.

requirements.txt
```diff
@@ -1,18 +1,17 @@
...
 flask-cors~=3.0.10
 flask-dropzone~=1.6.0
-flask-monitoringdashboard~=3.1.0
-flask-sqlalchemy~=2.5.0
-matplotlib==3.0.3
-mysqlclient~=1.4.6
-numpy>=1.16.4
-pillow>=7.1.0,<8.3.2
+flask-monitoringdashboard~=3.1.1
+flask-sqlalchemy~=2.5.1
```
willgraf, line 7: FYI I believe these patch upgrades are unnecessary since we use the ~=. It indicates that all 3.1.X are upgradable

.ebextensions/options.config
```diff
@@ -17,7 +17,8 @@ option_settings:
...
 
   aws:elasticbeanstalk:container:python:
-    WSGIPath: application.py
+    WSGIPath: application:application
 
-  aws:elasticbeanstalk:container:python:staticfiles:
-    /static/: deepcell_label/static/
+packages: 
+  yum:
+    mariadb-devel: []
```
willgraf, line 24: No longer necessary to provide the staticfiles path?

.github/workflows/tests.yaml
willgraf: I now see that 3.9 is not supported by EB. No problem.

.ebextensions/options.config
tddough98: The static files are now hosted in an S3 bucket instead of served by the flask backend

requirements.txt
tddough98: For flask-monitoringdashboard in particular, 3.1.1 is needed because 3.1.0 does not have Flask 2.0 support. Is it recommended to remove all patch upgrades and make all the dependencies like `example~=2.5.0`?

willgraf: Nope, no changes are needed, I just wanted to be clear.

.ebextensions/pip.config
tddough98: I got this from https://stackoverflow.com/questions/62457165/deploying-django-to-elastic-beanstalk-migrations-failed/63074781#63074781 and the answer claims that 

> I have verified with AWS Support that this is in fact the recommended way to do this. You must source the python environment, as with AL2 they use virtual environments in an effort to stay more consistent.

willgraf: OK sounds like it is indeed recommended.

.github/workflows/tests.yaml
tddough98: I'm open to adding 3.9 and 3.10 to the test action to check if they would pass and then keep them in if they do. If the tests need work, then let's push it off to another PR.

willgraf (approved):
LGTM

.github/workflows/tests.yaml
willgraf: Up to you, I like testing lots of versions but no EB support is a good reason not to include it.

tddough98: Eh I'll put it another PR let's just merge this in

PR #288, opened 2021-12-08
Add testing for python 3.9 and 3.10

tddough98:
In #287, we drop support for Python 3.6, but did not test for the newer Python releases 3.9 and 3.10. Here we add them to our test matrix to make sure our codebase works for these versions for when we'll need to switch to them down the road.

However, pytest-pep8 is incompatible with Python 3.10. Python 3.10 requires at least pytest 6.2.4 per [this issue](https://github.com/pytest-dev/pytest/issues/8539) and pytest-pep8 is incompatible with pytest 6, as seen in this error message
```
Direct construction of Pep8Item has been deprecated, please use Pep8Item.from_parent.
See https://docs.pytest.org/en/stable/deprecations.html#node-construction-changed-to-node-from-parent for more details.
```

As [pytest-pep8](https://pypi.org/project/pytest-pep8/) is not an actively mantained project and the last release was in 2014, I switched to [flake8](https://pypi.org/project/flake8/) and [pytest-flake8](https://pypi.org/project/pytest-flake8/). Flake8 checks for pep8 as well as pyflakes and circular complexity, keeping the style checks and adding some error checks.

In summary, this PR contains:

- additions to tests.yaml to test Python 3.9 and 3.10
- switches out pep8 checks for flake8 checks
- fixes issues highlighted by flake8

### DIFF
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -89,7 +89,7 @@ jobs:
 
       - name: Run PyTest
         run: |
-          pytest deepcell_label --cov deepcell_label --pep8
+          pytest deepcell_label --cov deepcell_label --flake8
 
       - name: Coveralls
         if: env.COVERALLS_REPO_TOKEN != null
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -60,7 +60,7 @@ jobs:
 
     strategy:
       matrix:
-        python-version: [3.7, 3.8]
+        python-version: [3.7, 3.8, 3.9, 3.10]
 
     steps:
       - uses: actions/checkout@v2
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -60,7 +60,7 @@ jobs:
 
     strategy:
       matrix:
-        python-version: [3.7, 3.8, 3.9, '3.10']
+        python-version: ['3.7', '3.8', '3.9', '3.10']
 
     steps:
       - uses: actions/checkout@v2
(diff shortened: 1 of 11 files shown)


.github/workflows/tests.yaml
```diff
@@ -60,7 +60,7 @@ jobs:
 
     strategy:
       matrix:
-        python-version: [3.7, 3.8]
+        python-version: [3.7, 3.8, 3.9, '3.10']
```
willgraf, line 63: Does 3.10 need quotes? I don't like that it looks different.

pytest.ini
```diff
@@ -25,7 +25,7 @@ norecursedirs=
 # E731 do not assign a lambda expression, use a def
 # W503 line break occurred before a binary operator
 
-pep8ignore=* E731
+flake8-ignore=* E731 W503
```
willgraf, line 28: nice catch.

willgraf (commented):
Looks good, just a couple questions/comments.

.github/workflows/tests.yaml
tddough98: It does need quotes, otherwise it's parsed as 3.1. You can see this in the actions for a0a77d6. I can add quotes to all of them if you'd prefer

willgraf: Yes, let's do that - the single value with quotes is bugging me. Thanks.

tddough98: Fixed

willgraf (approved):

PR #289, opened 2021-12-08
Add pre-commit hooks to automatically fix Python code

tddough98:
Here we enforce autoformatting of Python code with black. 

There are three contenders for python autoformatters: [autopep8](https://pypi.org/project/autopep8/), [yapf](https://github.com/google/yapf), and [black](https://black.readthedocs.io/en/stable/). Of these three, black is the most popular with [over 23k stars on github](https://github.com/psf/black). A number of high profile libraries are now using black, including Django, PyTest, tox, pandas, sqlalchemy, and virtualenv.

The main selling point of black is that it is very opinionated so that black makes all the decisions about code formatting. The only configuration I'm aware of is [max line length (default is 88)](https://black.readthedocs.io/en/stable/the_black_code_style/current_style.html#line-length), and [whether to replace single quotes with double quotes (default is to replace with double quotes)](https://github.com/psf/black/issues/118). I left both of these on their defaults.

I've also added an action to run black on pull requests to enforce its usage. The action checks that the code is consistent with black, and if not, it adds a commit to the PR to autoformat the code.


willgraf:
Let's use the single-quotes option so it better matches with the other lab projects.

tddough98:
I'm inclined to use black's opinion here. Here's their reasoning: [https://black.readthedocs.io/en/stable/the_black_code_style/current_style.html#strings](https://black.readthedocs.io/en/stable/the_black_code_style/current_style.html#strings)

To be clear, the `--skip-string-normalization` option does not replace double quotes with single quotes, it just doesn't replace single quotes with double quotes, which can leave inconsistent single and double quotes throughout the codebase. For example, `a, b = 'test', "test"` would be formatted to `a, b = "test", "test"` by default and would not change with `--skip-string-normalization`. If we start recommending black to contribute to the project, then using the default settings will add in double quotes that will not be caught or converted to single quotes.

tddough98:
[Black will not add an option for single quotes](https://github.com/psf/black/issues/594), so we'll have to add something like the [double-quote-string-fixer pre-commit hook](https://github.com/pre-commit/pre-commit-hooks/blob/master/README.md#double-quote-string-fixer) to enforce this.

willgraf:
Well...that's disappointing. Lots of good arguments for single quotes in the thread but I guess it's not going to happen. Personally, I agree that the double-quotes add a lot of visual noise and I do prefer single quotes.

Does the hook work? I don't want to overengineer it but I do have a preference.

tddough98:
Ok I got the precommits working. I added 

-`pre-commit` to requirements-test.txt
- `.pre-commit-config.yaml` to configure the pre commit hooks
- an action `precommithooks.yaml` to run the hooks on pull requests
 
We run `pre-commit install` to set up the pre-commit hooks locally in git.

tddough98:
I've switched to using the [git-auto-commit-action](https://github.com/stefanzweifel/git-auto-commit-action). I also added a personal access token to the [checkout action](https://github.com/actions/checkout) so tests are retriggered after an auto commit, [as indicated here.](https://github.com/stefanzweifel/git-auto-commit-action#commits-made-by-this-action-do-not-trigger-new-workflow-runs).

I removed the final newline from .flake8 in b034d32 to check that tests would run on the auto commit, and it seems like everything works as expected.

I don't think the drop in coverage is meaningful as it's only due to automatic reformatting.

Here's a summary of the pre-commit hooks that I set up:

- [black](https://github.com/psf/black) to replace the autoblack action that formats with black
- [isort](https://github.com/timothycrosley/isort) to sort imports
- [flake8](https://github.com/PyCQA/flake8) to correct flake8 issues, which includes [pycodestyle](https://github.com/PyCQA/pycodestyle) for pep8 checks, [pyflakes](https://github.com/PyCQA/pyflakes) for error checks, and [mccabe](https://github.com/PyCQA/mccabe) for complexity checks
- [pre-commit-hooks](https://github.com/pre-commit/pre-commit-hooks) for out-of-the-box hooks for pre-commit
  - trailing-whitespace to remove whitespace at the end of lines
  - end-of-file-fixer to check for a single newline that the end of files
  - debug-statements to check for debugger statements in the code
  - double-quote-string-fixer to replace double quoted strings with single quoted strings.
  
Here's a summary of the configuration changes to support these pre-commit hooks:
  - pyproject.toml to configure black and isort
  - .flake8 to configure flake8
  - .pre-commit-config.yaml to configure pre-commit
  - precommithooks.yaml to run the pre-commit hooks as an action on pull requests
  
Now that we check flake8 in pre commits, I've removed it from the testing workflow.

.github/workflows/precommithooks.yaml
```diff
@@ -0,0 +1,22 @@
...
+
+on: [pull_request]
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+        with:
+          token: ${{ secrets.PERSONAL_ACCESS_TOKEN }}
```
willgraf, line 11: Why is this needed? Can it not use the GITHUB_TOKEN that is provided by default?

.github/workflows/tests.yaml
```diff
@@ -89,7 +89,7 @@ jobs:
 
       - name: Run PyTest
         run: |
-          pytest deepcell_label --cov deepcell_label --flake8
+          pytest deepcell_label --cov deepcell_label
```
willgraf, line 92: Clarification: No need to use --flake8 because it is linted and commited before this runs?

willgraf (approved):
Great improvement! We should consider integrating this into the rest of our `deepcell` projects as well.

.github/workflows/tests.yaml
tddough98: Flake8 now runs in parallel in the pre-commit hook action.

If the author happens to have used pre-commit to set up the pre-commit hooks with `precommit install` then the pre-commit hook does nothing, but if they have not installed the hooks, then the action runs and commits the hooks for them.

tddough98:
For sure! I'm happy to help out or code review adding these pre-commit hooks to any other repos, just tag me and I'll take a look

PR #290, opened 2021-12-09
Add prettier pre commit hook

tddough98:
I added our prettier setup for formatting the javascript side of the project to the pre-commit configuration and action. I also updated the configuration to be closer to prettier's recommended style, only overriding to use single quotes instead of double quotes as we do for the Python formatting as well. Specifically, I removed `"semi": true,` from .prettierrc because that's already the prettier default and I removed `"arrowParens": "avoid",` after reading [prettier's reasoning here](https://prettier.io/docs/en/options.html#arrow-function-parentheses).

I ran the updated prettier config on all files, which is why there are so many changed files. The only changes of note are in .prettierrc and .pre-commit-config.yaml.


willgraf (approved):
🚀

Issue #274, opened 2021-11-02
Label highlighting in chrome covers the raw image


tddough98:
**Describe the bug**
Chrome changed its behavior on how to display stacked canvas, causing selected labels to be filled with opaque white.

**To Reproduce**
Select a label by clicking on it. When selected as the foreground label, the label is filled in with an opaque white instead of the translucent white that used to appear. Firefox still shows the translucent highlight.

This occurs on Google Chrome Version 96.0.4664.27 (Official Build) beta (x86_64).

**Expected behavior**
The highlight should be translucent so the raw image is still visible.

**Screenshots**
<img width="1084" alt="Screen Shot 2021-11-02 at 4 44 00 PM" src="https://user-images.githubusercontent.com/40068392/139966787-76f808a3-bce2-4cff-b48a-05d70a09b2d8.png">

Issue #273, opened 2021-11-02
Fix canvas updates when adding or removing channels


tddough98:
**Describe the bug**
The canvas does not update immediately when adding and removing channels in multi-channel mode.

**To Reproduce**
On a projet on label.deepcell.org, switch to multi-channel mode and press the Add Channel button or remove a channel with the pop-up menu. The canvas does not update right away and instead requires another interaction like moving the mouse over the canvas.

**Expected behavior**
The canvas should update as soon as a channel is added or removed.

PR #294, opened 2022-01-07
Accelerate image processing on the client with GPU.js

tddough98:
Here, we improve the image processing and display for the DeepCell Label client with GPU acceleration. I revamped existing `____Canvas`components to use a kernel compiled with [GPU.js](https://gpu.rocks/#/) instead of processing image data on the CPU, resulting in large speedups for image processing, especially for larger images. For a 4.5 megapixel image, changing the range of a channel now take around 6ms to complete the render compared to about 80ms in our existing deployment.

Each Canvas is now explicitly composed into a single canvas, instead of implicitly composed with a stack of canvases in the same position on the page. This address issue #274.

I've also addressed #273 by sending some additional `ADD_LAYER` and `REMOVE_LAYER` events to parent actors to ensure that the relevant hooks update when the event occurs.

I removed routes and logic from the backend that are now exclusively handled by the client including

- creating a label image from the label array with a colormap
- creating outlines from the label array


deepcell_label/blueprints.py
```diff
@@ -94,24 +93,13 @@ def add_raw(token):
...
-
-
 @bp.route('/api/array/<token>/<int:feature>/<int:frame>')
 def array(token, feature, frame):
     """ """
     project = Project.get(token)
     if not project:
         return jsonify({'error': f'project {token} not found'}), 404
-    labeled_array = project.get_labeled_array(feature, frame)
+    labeled_array = project.get_labeled_array(feature, frame).flatten()
```
willgraf, line 102: Why is this change required?

deepcell_label/imgutils.py
```diff
@@ -1,26 +1,9 @@
...
-        # apply the colormap
-        imgarr = cmap(imgarr, bytes=True)
-
-    img = Image.fromarray(imgarr)
-    img.save(out, format='png')
-    out.seek(0)
-    return out
 
 
 def grayscale_pngify(imgarr):
```
willgraf, line 9: Can this one be computed on the frontend as well?

deepcell_label/models.py
```diff
@@ -117,7 +116,6 @@ class Project(db.Model):
     channel = db.Column(db.Integer, default=0)
     feature = db.Column(db.Integer, default=0)
     scale_factor = db.Column(db.Float, default=1)
-    colormap = db.Column(db.PickleType)
```
willgraf, line 119: Does removing fields cause any issues with the schema difference?

Or does it just get ignored bc SQL doesn't include the field in the queries?

visualizer/src/Canvas/Labeled/LabeledCanvas.js
```diff
@@ -1,104 +1,66 @@
...
   const canvasRef = useRef();
-  const ctx = useRef();
-  const hiddenCanvasRef = useRef();
-  const hiddenCtx = useRef();
 
   useEffect(() => {
-    ctx.current = canvasRef.current.getContext('2d');
-    ctx.current.imageSmoothingEnabled = false;
+    const canvas = canvasRef.current;
+    const gl = canvas.getContext('webgl2', { premultipliedAlpha: false });
```
willgraf, line 33: What does the webgl2 context mean?

visualizer/src/Canvas/Labeled/LabeledCanvas.js
```diff
@@ -1,104 +1,66 @@
...
-    opacity,
-    sw,
-    sh,
-    sx,
-    sy,
-    zoom,
-    width,
-    height,
-  ]);
+    kernel.current(labeledArray, colormap, foreground, highlight, highlightColor, opacity);
```
willgraf, line 59: Could you please add some brief comments explaining what the effect is supposed to do?

willgraf (changes requested):
As discussed, I think this is a good opportunity to add some tests to the Canvas components, especially as we have moved some more complicated logic into them.

It looks as though there is a failure on the 17.x node version for tests as well.

deepcell_label/models.py
tddough98: Dropping a column should be fine. The old column will still be in our production database, but it will be ignored and unused. Old rows will still have data in the column while new rows will not write to the column, and we can't access the column through SQLAlchemy because it's no longer in the schema

visualizer/src/Canvas/Labeled/LabeledCanvas.js
tddough98: There are several context types available for a canvas, including '2d', 'webgl', and 'webgl2'([https://developer.mozilla.org/en-US/docs/Web/API/HTMLCanvasElement/getContext](https://developer.mozilla.org/en-US/docs/Web/API/HTMLCanvasElement/getContext))

This line specifies the premultipliedAlpha setting for the webgl2 context of the canvas. We need it for the LabeledCanvas, OutlineCanvas, BrushCanvas, and ThresholdCanvas as they can include transparent/translucent pixels. Without this setting, composing the canvases block the raw image underneath. It's after this example in the gpu.js documentation: [https://github.com/gpujs/gpu.js/#alpha](https://github.com/gpujs/gpu.js/#alpha)

visualizer/src/Canvas/Labeled/LabeledCanvas.js
tddough98: The `kernel.current(...)` line rerenders the canvas powered by GPU.js for the current component. The next line `setCanvases(...)` rerenders the parent canvas with the updated canvas for this component. I added these comments to these effects in each Canvas component that uses it.

deepcell_label/imgutils.py
tddough98: Yes, I updated the backend to send the raw data in an array just like the labeled array and then computes the image on the client.

deepcell_label/blueprints.py
tddough98: I made this change to work better with GPU.js, but I'm not 100% sure it's needed... let me experiment a bit and see if it's kosher to pass a 2D array to GPU.js, and if it is, I'll prefer that and send the unflattened arrays.

deepcell_label/models.py
```diff
@@ -446,42 +438,24 @@ def make_first_payload(self):
...
-        label_png = pngify(
-            imgarr=np.ma.masked_equal(label_arr, 0),
-            vmin=0,
-            vmax=self.get_max_label(feature),
-            cmap=self.colormap,
-        )
-        return label_png
+        array = raw_frame.frame[..., channel]
+        normalized = (array - np.min(array)) / (np.max(array) - np.min(array)) * 255
+        print(np.min(array), np.max(array), np.max(normalized), np.min(normalized))
```
willgraf, line 449: Probably ought to remove this print/debugging line before merging.

deepcell_label/blueprints.py
tddough98: Got it to work, so it's switched back now to sending 2D arrays instead of flattening into a 1D array

willgraf (changes requested):
Thanks for the changes, everything looks pretty good - just found one spot with an extra print statement that is probably unnecessary.

willgraf (approved):
LGTM. Some unit tests for the individual canvas components would be nice to include though.

tddough98:
I added a test that Canvas sends events to the canvasMachine and that BrushCanvas has a canvas of the right size. I spent some time trying to test the canvas contents of BrushCanvas, but I hit some walls. As Jest runs tests in a node environment, some browser APIs, including the canvas, are not available. I installed [canvas](https://www.npmjs.com/package/canvas) as a dev dependency which recreates the canvas API for node, but this does not include webgl functionality, which is a key need here. One solution seemed to be running tests with [jest-electron](https://www.npmjs.com/package/jest-electron) so that webgl is available, but that will require ejecting create-react-app, which I'm not inclined to do in this PR. I'll explore that more another time, but for now, the two components tested should provide at least some foundation and models for testing more canvas functionality in the future.

tddough98:
Unfortunately another bamboozling update for testing components with GPU.js. The standard tool for javascript code coverage is istanbul, which turns out is incompatible with GPU.js [https://github.com/gpujs/gpu.js/issues/663](https://github.com/gpujs/gpu.js/issues/663). The recommendation is to use v8 code coverage instead, but this requires using the coverageProvider option for jest, which is not supported with create-react-app...

I'm taking out the BrushCanvas test

PR #296, opened 2022-01-13
Fix GPU.js handling of WebGL

tddough98:
With the most recent pull request #294, I noticed that Firefox and Chrome handled the new canvas as expected, but Safari was not behaving correctly. It seems that that Safari only received WebGL2 support in Safari 15 released a few months ago with macOS 15: [https://caniuse.com/webgl2](https://caniuse.com/webgl2).

I fixed the initialization of the GPU-accelerated contexts to check which context is available in the browser and set one or the other up as needed. I also had issues with the WebGL canvases clearing their data after being drawn, causing it to appear once and then to appear blank in later renders. To fix this, I drew the WebGL kernel output onto a separate canvas as soon as it is created, and this other canvas with a standard 2d context does not get cleared when drawn. This strategy works just as well for WebGL2, WebGL and canvas based kernels, so I implemented it as the standard approach regardless of WebGL support

### DIFF
--- a/visualizer/src/Canvas/ComposeCanvases.js
+++ b/visualizer/src/Canvas/ComposeCanvases.js
@@ -1,7 +1,7 @@
 import { makeStyles } from '@material-ui/core';
 import { useSelector } from '@xstate/react';
 import React, { useEffect, useRef } from 'react';
-import { useCanvas } from '../ProjectContext';
+import { useCanvas, useDrawCanvas } from '../ProjectContext';
 
 const useStyles = makeStyles({
   canvas: {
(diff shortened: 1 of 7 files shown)


willgraf (approved):
LGTM, though I'm not very knowledgable on the webGL stuff.

One thought, it might be nice to make a helper to re-use some of this gl/gl2 logic (if possible).

tddough98:
> One thought, it might be nice to make a helper to re-use some of this gl/gl2 logic (if possible).

I pulled out both the logic to setup the context for the kernel and the separate canvas to draw on into custom hooks in `ProjectContext`

PR #297, opened 2022-01-13
Fixes for tracking before release

tddough98:
This adds in a couple of fixes, mostly for tracking jobs, including

- enabling uploading .trk from the homepage
- removing an extra FrameSlider above the image controls
- hiding the Frame label when FrameSlider is hidden
- fixing the colors provided to the Division components to be provided in hex format instead of a list of [r, g, b]


willgraf (approved):
LGTM, I really like the useHexColorMap hook! nice cleanup.

PR #299, opened 2022-01-14
Wrap GPU.js kernel functions in templates

tddough98:
Our minified production code replaces expression like `if (x) { y }` to `x && y`, breaking how GPU.js compiled functions into a kernel. By wrapping the functions in ticks, they won't be minified and all the syntax is preserved for GPU.js when it compiles the function at runtime.

These issues on the GPU.js repo detail this: 
[https://github.com/gpujs/gpu.js/issues/152 ](https://github.com/gpujs/gpu.js/issues/152) 
[https://github.com/gpujs/gpu.js/issues/584](https://github.com/gpujs/gpu.js/issues/584)


willgraf (approved):
What a strange issue.

Wrapping the lines in ticks is a multiline/template string, I think. I don't _love_ passing a string of javascript code to the kernel but if it solves the issue I'm happy.

Consider adding TODO in there to research how to replace with minifify-able code?

PR #300, opened 2022-01-18
Fix minified brush kernel function

tddough98:
Similar to #299, the minified kernel functions for the brush canvas was causing breaking changes for the kernel compilation. This quick fix moves the brush kernel function into a template string and moves a custom function into a local declaration so that it cannot get renamed during minification. 

[label.deepcell.org](label.deepcell.org) has been updated with this fix and it can be tested there


MekWarrior:
Given the similarity to #299, we should add a TODO here as well. It may be worth adding an issue for both TODOs to further investigate the root cause of the minification issue.

MekWarrior (approved):
Outside of the issue/todo this retains the pattern of an earlier PR and should be fine to merge.

PR #301, opened 2022-01-25
Fix quality control canvases  

tddough98:
#294 added hooks `useAlphaKernelCanvas` and `useDrawCanvas` to encapsulate the logic to set up canvases. When doing QC with DCL (switching between multiple projects on the same page), these canvases were not switched out properly, leaving the canvas for the previous project on screen.

Here, we add the top level project machine to these dependencies of these hooks, so when the project switches, the canvases will be set up properly for the new project.


MekWarrior:
Do we need a corresponding change for the `LabelCanvas` or `OutlinedCanvas`?

tddough98:
Good question, I thought LabelCanvas and OutlinedCanvas would behave fine because they don't modify the canvas behavior themselves like how RGBCanvas sets the globalCompositeOperation. I created two projects from the same file and added a brush stroke in one so I could tell them apart, and when I tested switching between two (nearly) identical projects, the canvases were not updating correctly.

I've moved from returning refs to canvases from `useDrawCanvas` and `useAlphaKernelCanvas` to returning a canvas directly. [Refs in React](https://reactjs.org/docs/refs-and-the-dom.html) are often used to access DOM nodes and they allow persistent values between renders, but they add complexity to the code and given that rerendering the components do not rerun the hooks, it is just as performant to directly return the value from the hook.

With these fixes, even switching between identical project works and they each have separate state that we can switch back and forth between.

tddough98:
@MekWarrior I still need approval to merge this in. I think it should be set to go, but let me know if you have any questions or concerns

MekWarrior (approved):
Looks good. Thanks for the changes and the test.

PR #302, opened 2022-01-27
Use event buses to facilitate cross-actor message passing

tddough98:
This PR adds event buses to coordinate message passing between actors. Previously, message passing between siblings in the actor tree had to pass events up to the parent and the parent had to forward events to the appropriate children.

Event bus make message passing more flexible. They allow actors to broadcast events without knowing who need to receive the event and they enable two way communication.

This change declutters the message passing management previously in the root `projectMachine` and opens up future flexibility for message passing. We'll need to refine message passing as we move forward with the new label format, but the event bus pattern should be a good starting point.

The EventBus and fromEventBus source code and pattern is taken from [xstate-behaviors](https://github.com/ChrisShank/xstate-behaviors/blob/master/src/event-bus/from-event-bus.ts), which recommends 

> This package is not yet release as it is in a very early alpha state. If you would like to use any of these helpers now its recommended to copy and paste them into your codebase for the time being!

Here are the state machines with an event bus, which events they send to the buses, and which machines use those events.

- image 
  - sends `FRAME`, `FEATURE`, and `CHANNEL` events
  - `apiMachine` uses these events append to outgoing EDIT events so the backend edit the correct image slice
- labeled
  - sends `LABELED_ARRAY` and `LABELS` events
  - `canvas` uses `LABELED_ARRAY` to find the label under the cursor
  - `select` uses `LABELS` to cycle through labels
- canvas 
  - sends `HOVERING`, `COORDINATES`, `mouseup`, and `mousedown` events
  - `toolMachine` forwards mouseup and mousedown events to the active tool
  -  tools (brush, threshold, flood, etc.) use HOVERING and COORDINATES to track the cursor
- select 
    - sends `FOREGROUND`, `BACKGROUND`, and `SELECTED` events
    - tools use these to know what labels to edit
- raw 
  - sends `GRAYSCALE` and `COLOR` events
  - `toolMachine` uses these to activate or deactivate grayscale specific tools (like threshold, autofit, and watershed)
- api
  - sends `EDITED` events to labeled, so labeled knows to update it's data after an edit
  - tool machines send `EDIT` events to initiate an edit on the backend

The undo eventBus is an exception. The `undoMachine` does not send events to other actors on the event bus. Instead, the bus allows other machines to register actors in the undo/redo history with `ADD_ACTOR` events.

Here are some of places where I expect event buses will be useful for the new label format

- defining additional label types
  - event buses let new labels to send & receive message to existing actors without the existing actors needing to know about the new label type
  - when editing one label, dependent labels need to receive the edited labels to know how to update themselves. An event bus lets the edited label broadcast changes to all actors listening to the bus
  - if a new label needs application state like the canvas position, the new labels can listen to the canvas bus without the canvas actor needing to know
- undo/redo
  - undo/redo needs to capture all the changes that propagate from a single label change. If all changes are passed on the same event bus, undo/redo can listen to the event bus and know all the labels that changed
  - undo/redo needs to restore UI state across many actors. An event bus could ensure all actors that saved their state to be undone/redo receive messages to restore their state


msschwartz21 (commented):
Can you comment on the utilization of typescript for defining the `EventBus` class, e.g. what assertions/validations are enforced?

tddough98:
Most of the types used with EventBus are imported from the state management library we're using XState, which checks the shape of data coming from state machines, like the context that they store and the events that they send. These are the most generic types possible for XState, and to use them more effectively, we'll have to extend them in our code base to the specific contexts and events that we use for our state machines. We see this in the new type defined things that receive events `type Subscriber<TEvent extends EventObject> = (event: Event<TEvent>) => void;`, where we need to create specific `TEvent` types for our events that extend the base `EventObject` type from XState.

PR #303, opened 2022-02-01
Remove node 12 from testing matrix

tddough98:
Our frontend testing suite has started failing on the yarn install step because Node 12 has reached its end of life. This PR drops node 12 from the testing matrix.

### DIFF
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -11,7 +11,7 @@ jobs:
 
     strategy:
       matrix:
-        node-version: [12.x, 14.x, 16.x, 17.x]
+        node-version: [14.x, 16.x, 17.x]
 
     steps:
       - uses: actions/checkout@v1


tddough98:
Here is the node EOL schedule for reference: [https://endoflife.date/nodejs](https://endoflife.date/nodejs)

msschwartz21:
Just a question out of curiosity, is there anywhere else in the project where the node dependency/requirements are described?

msschwartz21 (approved):

tddough98:
Node is only used for testing. Node is typically used to run javascript on a server, like the test suite that's running on GitHub actions. It's the same V8 engine that runs in Chrome, so it provides a good proxy for how javascript will run in the browser, but it's still just a proxy. For example, the canvas API is not available in node as node doesn't have a rendering engine. Given that we have a fair amount of canvas logic, I've explored other testing options like using [jest-electron](https://www.npmjs.com/package/jest-electron) (see the discussion for #294), which runs tests with both node and a rendering engine to make it closer to a real browser, but that involves ejecting all the tooling that create-react-app manages.

Having a complete testing matrix for node is probably overkill because this isn't really a node project, so if there's ever an issue with Label tests eating too much GitHub Actions minutes, we should look into cutting down the testing matrix to just the current version.

Issue #305, opened 2022-02-03
DeepCell-Label online service does not read *.npz files


kim-fehl:
**Describe the bug**
Neither [demo file](https://caliban-input.s3.us-east-2.amazonaws.com/janelia_demo/HeLa-S3_janelia_demo_version.npz) nor [ours](https://github.com/vanvalenlab/deepcell-label/files/7994506/im11_2.zip) work. 

**To Reproduce**

1. Open https://label.deepcell.org/
2. Drag and drop *.npz file
3. Press Upload button 
4. See error text: "undefined"

**Expected behavior**
File being uploaded and web interface opened

**Desktop (please complete the following information):**

- OS:Ubuntu 20.04
- Browser: Google Chrome Version 97.0.4692.99 (64-bit)
![dcl_webservice_error_2022-02-03_15-30](https://user-images.githubusercontent.com/25846538/152343840-2cf7ced4-4a3d-4d4d-8e58-dec922545555.png)

tddough98:
Thanks for raising this issue. A 413 error is from an attached file that is too large, and it looks like some of the changes we've made to our server has changed the default configuration making so even our own demo files are too large to load.  I've updated the configuration to allow uploads of up to 50 megabytes in. [label.deepcell.org](label.deepcell.org) has already these changes online so you should be able to use the site with your files now

I checked that both our [demo file](https://caliban-input.s3.us-east-2.amazonaws.com/janelia_demo/HeLa-S3_janelia_demo_version.npz) and [your file](https://github.com/vanvalenlab/deepcell-label/files/7994506/im11_2.zip) load as expected. I noticed your npz file has channels first, so make sure you pick the CZYX dimension order when uploading the file so it displays properly. If you have any more issues or questions, don't hesitate to ask and I'll be glad to help out!

PR #306, opened 2022-02-03
Set file upload size limit to 50 megabytes

tddough98:
As reported in #305, drag and drop is not working on the label.deepcell.org homepage due to the 413 errors reporting that attached files are too large.

I believe with our update to Amazon Linux 2 for the backend in #287, the behavior for nginx changed. With this new platform, the configuration looks a bit different/ With the previous platform, we would [set up this file with a `.config` file ](https://stackoverflow.com/questions/18908426/increasing-client-max-body-size-in-nginx-conf-on-aws-elastic-beanstalk) but with Amazon Linux 2, [this approach no longer works ](https://stackoverflow.com/a/70959127/14458070) and instead we put the conf file in the `.platform` folder.

### DIFF
--- a/.ebextensions/files.config
+++ b/.ebextensions/files.config
@@ -1,0 +1,7 @@
+files:
+  "/etc/nginx/conf.d/proxy.conf" :
+    mode: "000755"
+    owner: root
+    group: root
+    content: |
+      client_max_body_size 50M;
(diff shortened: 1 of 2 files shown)


msschwartz21 (approved):
Thanks for the quick fix!

PR #307, opened 2022-02-04
Add yarn.lock

tddough98:
Yesterday I helped @msschwartz21 debug some issues with running Label locally, and I believe the issue came down to dependency mismatches. I've learned that for applications like Label, it is best practice to track yarn.lock in the repo as it ensures the exact version of dependencies are installed across machines.

Here's a stack overflow answer confirming that yarn.lock should be tracked for applications: https://stackoverflow.com/a/40206145/14458070

Here's an issue on the yarn repo as well: https://github.com/yarnpkg/yarn/issues/1583


msschwartz21:
Does introducing the lock file change the installation commands that should be used for local deployment? If so, can we make a note in the README so that these details don't get lost.

tddough98:
Adding yarn.lock doesn't change any commands for local deployment. The frontend is still started with `yarn start`. Having a `yarn.lock` means that `yarn install` (or just `yarn`) installs the exact versions of each dependency across all machines.

Having a `yarn.lock` does mean that users should not use npm instead of yarn, but we don't mention using `npm` anywhere at the moment.

I wonder if we could use a separate readme about setting up dependencies/environments for the first time, and then only mention the 

This readme could cover per machine setup like installing python and node and per repo setup, like

- creating a virtual python environment
- installing python dependencies
- installing javascript dependencies with yarn

Once these are completed once, users will only have to run the repeatable `flask run` + `yarn start` commands.

msschwartz21 (approved):

MekWarrior (approved):

PR #309, opened 2022-02-15
Add separate README with instructions for first time setup

tddough98:
I've added a separate README with detailed instructions on setting up Python and Node for first-time users. This adds details for users who have not used Python or Node before without adding more to the main README that is not as useful for experienced or repeat users. I also tweaked the language and order of some of the README.


FIRST_TIME_SETUP.md
```diff
@@ -0,0 +1,70 @@
...
+
+## Setup Node.js (Javascript)
+
+As DeepCell Label is a web application, the frontend is written in JavaScript. We'll install Node.js, a server-side Javascript platform, and Yarn, a Javascript package manager.
+
+<!-- adapted from https://gist.github.com/Igormandello/57d57ee9a9f32a5414009cbe191db432 -->
+
+### Node
+
+- #### Node installation on Windows and macOS
```
MekWarrior, line 37: I found this a little disorienting because the instructions for the backend seem to be targeted at Mac/Ubuntu users. If we include Windows instructions, it might be best to test on that OS or have a disclaimer stating that it was designed for Mac/Ubuntu but should run fine multiple OS.

MekWarrior (approved):
Nice changes. Only one small comment based on something that has caught us out in the past, but it's very minor.

PR #311, opened 2022-02-15
Load project data in bulk and decouple label image editing from database

tddough98:
This PR is a two-in-one. Earlier work on the PR focuses on how data loading into client. Previously, we loaded each frame in series and we now load all data for a project in bulk at the start of the application.

Loading data in bulk aligns with the longer term goal of decoupling the client from the server and removes a significant amount of complexity from state machines that used to load data (see `imageMachine` and its child/grandchild actors). Instead, we load and manage all array data in the `arrayMachine` and component that need the data access it through `useArrays`. We also introduce `labelsMachine` to manage the cell info, which will be the footprint for the state machines for the new label format. We don't use the new label format yet in this PR, but rather lay the groundwork to substitute the prototype state machines from other branches for the new label format.

Later work on the PR focuses on decoupling the Edit module on the server from the database, which also aligns with decoupling the client and server. Instead of telling the server which image to read from the database, we send the image itself and get back the updated image. By loading the array data in bulk, we can treat the client copy of the project data as the ground truth, so to edit and iterate on the labels, we need to send the current labels to an editing API and save the updates on the client.

This PR does break some key features, namely undo/redo and tracking. To return to feature parity, we'll need to

- persist label image snapshots in IndexedDB on the client
- reimplement tracking operations for the new label format

Each of these are significant development efforts that deserve dedicated effort and review, so I think it's better to review and merge this PR into the develop branch now and add back these features piecewise. I also want to call out some new directions/features that are highlighted/enabled by this PR

- separating API calls from the state management tree so we can easily test the state tree by mocking the API
- running tests with [jest-electron](https://github.com/hustcc/jest-electron)
- improving the label image editing API by sending and receiving zip files with [zip.js](https://gildas-lormeau.github.io/zip.js/)


tddough98:
Because of the order #313 and #314 were merged into this branch, there are changes in this PR that have already been reviewed.

These files have already reviewed

- .github/workflows/tests.yaml
- .gitignore
- all cypress and javascript testing related files

Files with core changes for this PR include
- deepcell_label/blueprints.py: serves array buffers instead of preprocessed images and loads image data from the client in the edit route
- deepcell_label/label.py: uses the image data sent by the client instead of reading from the database\
- visualizer/src/service/apiMachine.js: sending label and raw images to the edit route
- visualizer/src/App.js: uses loadMachine to get project data and provides it to projectMachine
- visualizer/src/service/loadMachine.ts: loads project data
- visualizer/src/service/arraysMachine.js: manage the image arrays
- visualizer/src/service/fetch.js: gets and processes array data from the server
- visualizer/src/service/imageMachine.js and rawMachine.js and labeledMachine.js and channelMachine.js and featureMachine.js : removes loading logic
- visualizer/src/service/labelsMachine.js: manages label data other than image arrays

Other smaller changes include
- renaming `LOAD___` events to `SET___` events
- getting array data in components with useArrays instead of useRaw and useLabeled
- no longer checking if raw or labeled machines exist before rendering a component

msschwartz21:
> * no longer checking if raw or labeled machines exist before rendering a component

Is this change made possible because the all of the data is loaded at the start of the application instead of retrieving as needed from the database?

tddough98:
> > no longer checking if raw or labeled machines exist before rendering a component
> 
> 
> Is this change made possible because the all of the data is loaded at the start of the application instead of retrieving as needed from the database?

previously, the raw and labeled machines were instantiated once the data was loaded. Now they are instantiated with the rest of the state machines and have a `waiting` state until they receive their data in a `LOADED` event

msschwartz21 (approved):
Thanks for being so patient waiting for my review! I think I am slowly but surely getting a better sense of how all these pieces fit together although I wish I had more feedback for you at the moment. 

I was wondering if you could briefly address how these changes impact the maximum data size that the application can handle, if at all?

tddough98:
This PR should not affect maximum data size handled as the data still has the same memory footprint. Previous versions load and stores the same amount of image data, but loaded each frame sequentially.

I'm not sure what bottleneck currently limits image sizes, but some possibilities are

1. sending large over the network when editing labels (i.e. editing latency becomes too high)
2. GPU.js code that processes images cannot handle too large images
3. images are larger than the available RAM

I expect that 1 would be the first bottleneck, then 2, then 3 because broadly speaking network bandwidth << GPU memory < RAM. I'm currently working on sending crops/subsets of images & labels to be edited by the server to decouple editing latency from image size. To deal with 2 and 3, we'll have to use new strategies like [viv](https://github.com/hms-dbmi/viv) such as multi-resolution image pyramids that load low resolution copies quickly then fill in with more detail after zooming in and splitting images into tiles and loading each tile separately. I think avivator also uses these strategies for GPU based display.

PR #313, opened 2022-03-02
Test React component tree and state management by mocking load events

tddough98:
Here I add a separate loadMachine to fetch data from the Flask backend. We separately create a state management tree in a waiting state and then send a LOADED event with project data once the loadMachine reaches its final state. By separating the API calls from the rest of the state management, we can adds tests that covers the full component tree by sending a mocked LOADED event to the state tree.

The end-to-end tests introduced in #314 provide a similar level of test coverage, with the added benefit of running the backend server as well and running in a real browser. Here, the tests are run by jest instead of cypress, which makes them more lightweight and easy to run often.

It's worth keeping both testing approaches in the codebase so we have the option to choose which we want to run, even if the Cypress tests are more powerful overall.


msschwartz21 (approved):
I'm going to summarize what I see as the big changes just to confirm my own understanding. By creating a new state machine for loading data, you are also able to set up a corresponding event bus that contains all of the relevant attributes for the data. This means that in downstream code you can retrieve information like width, height, channels, etc from the event bus instead of manually passing each piece for information between functions. Additionally the downstream code will not run until the LoadMachine updates its state to "loaded".

The end result of all of these changes is that you can easily mock out the process of loading data in order to test other pieces of the code quickly and independently.

PR #314, opened 2022-03-02
End to end testing with Cypress

tddough98:
Here I add an e2e-tests step the the tests workflow that runs end-to-end tests with [Cypress](cypress.io).

There are two true end-to-end tests, one for a tissue segmentation job and one for a tracking job, that picks an example file from the homepage, download the files on the backend, load them into the database, and serve them to the frontend, providing high value confidence that the full user workflows will work in production. True end-to-end tests are slower more expensive to run, so other tests instead use stubbed data to quickly load and interact with a page. We have near complete coverage of the React components, while most of the uncovered code is interactions defined in state machines. Future work can cover this code by interacting with components that dispatch events to state machines.

We measure code coverage on the frontend for these end-to-end tests and add that coverage to coveralls. At the moment, the Cypress tests do _not_ capture backend code coverage. I started to explore running the flask backend with [coverage](coverage.readthedocs.io). The flask backend is a long running process which needs to be terminated gracefully with SIGTERM to generate a .coverage file (https://coverage.readthedocs.io/en/latest/subprocess.html). As GitHub Actions manages the processes in each step, and the [Cypress GHA](https://github.com/marketplace/actions/cypress-io) runs the flask backend under, I couldn't find a way to send SIGTERM to the process before the Cypress step ended and terminated the process itself without generating a .coverage file. For now, we're losing out on measuring this backend coverage, but the end-to-end testing value is still as tests will fail if there are issues in the covered backend code.


msschwartz21 (approved):
Can you clarify why there are some changes to files in visualizer/src?

Otherwise this looks like a really great enhancement to our testing routine!

tddough98:
The changes in the src folder include

- App.js: fixing a warning that the machine passed to useMachine changes between renders
- Canvas.js: update name of state to fix the loading wheel
- LayerOptions.js: add a data-testid to make a component easier for tests to access
- apiMachine.js: add missing actions
- brushMachine.js: a bug fix when using the brush before moving the mouse over the canvas

PR #318, opened 2022-04-27
Add imagecodecs to requirements

tddough98:
I have been seeing this error come up in the backend logs when users try to load tiffs from the Label homepage.
```
ValueError: <COMPRESSION.LZW: 5> requires the 'imagecodecs' package
```

This adds imagecodecs to the requirements for the backend. It also drops Python 3.7 from our testing matrix as the latest release of imagecodecs is for Python 3.8+ per this NEP (like PEP for numpy) https://numpy.org/neps/nep-0029-deprecation_policy.html. We now use Python 3.8 for hosting on Elastic Beanstalk, so this matches our use.

### DIFF
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -24,7 +24,7 @@ jobs:
       - name: Cache node_modules
         uses: actions/cache@v2
         with:
-          path: 'visualizer/**/node_modules'
+          path: "visualizer/**/node_modules"
           key: ${{ runner.os }}-modules-${{ hashFiles('visualizer/**/yarn.lock') }}
           restore-keys: |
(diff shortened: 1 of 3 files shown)


MekWarrior:
I have run across this error in benchmarking locally as well. I believe it is a useful addition, especially for something user-facing like DCL.

msschwartz21 (commented):
Seems like a good change! Happy to approve once the pre-commit workflow is passing.

tddough98:
Bumped the black version in the pre-commit config to fix the Action. Here's a stack overflow post about the error: https://stackoverflow.com/questions/71673404/importerror-cannot-import-name-unicodefun-from-click

msschwartz21 (approved):

PR #319, opened 2022-04-27
Handle batches when loading TIFFs and zips of TIFFs

tddough98:
As we support batch dimensions on the kiosk-frontend, we need a way to handle differently shaped input images and predicted labels from the predict page.

 - Adds back a dimension order for the image file
 - Checks if tiff filenames contain `batch_XX` when loading a zip of tiffs and groups and loads files by their batch number


msschwartz21 (approved):

tddough98:
Added a test for loading multiple batches and fixed the coverage measurement with cypress tests by updating @cypress/code-coverage

PR #324, opened 2022-05-11
Send zip to edit route

tddough98:
Here we replace the form with multiple files sent to edit route with a single zip file of all the data created with [zip.js](https://gildas-lormeau.github.io/zip.js/). This compresses the data on the client before sending it over the network, reducing the network bandwidth. This also provides an extensible way to edit more labels on the server by adding another file to the zip sent to and from the server.


deepcell_label/label_test.py
```diff
@@ -138,65 +187,93 @@ def test_action_active_contour_label_too_large(self, app):
...
+                overlaps=overlaps,
+                action='dilate',
+                args={'label': 1},
+                write_mode='overwrite',
+            )
+            np.testing.assert_array_equal(edit.labels, expected_labels)
 
     def test_action_dilate_exclude(self, app):
-        """Tests that a dilated label does not affect other labels when write_mode is exclude."""
+        """Filated label does not affect other labels when write_mode is exclude."""
```
msschwartz21, line 266:
```suggestion
        """Dilating label does not affect other labels when write_mode is exclude."""
```

msschwartz21 (approved):
LGTM!

PR #327, opened 2022-05-17
Export edited labels

tddough98:
This PR closes the loop for editing labels on the dev branch by letting us exporting edited data. It provides two routes `/api/download` to download the edited labels to the users computer and `/api/upload` to upload them to an S3 bucket.

As the frontend cannot easily create a TIFF file, we send a zip with binary array data to the server which repackages these binary files into OME TIFF files that are more user friendly and compatible with other tools like Fiji.

Now that all the pieces are in place for a full data annotation cycle, I will put a demo instance online to test out all the new features. I still need to work more on editing the lineage on this branch before we can deploy a tracking job. Tracking labels can be loaded and displayed, but the editing is not functional yet.

### DIFF
--- a/deepcell_label/blueprints.py
+++ b/deepcell_label/blueprints.py
@@ -11,6 +11,8 @@ import requests
 from flask import Blueprint, abort, current_app, jsonify, request, send_file
 from werkzeug.exceptions import HTTPException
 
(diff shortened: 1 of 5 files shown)


msschwartz21 (approved):
The download button is not working on the current label-dev.deepcell.org, but assuming that it is a bug that will be corrected in one of the next PRs, then I'm happy to approve this PR.

PR #330, opened 2022-05-28
Create cells labels

tddough98:
This replaces the "overlap matrix" with a cells.json file of cell labels. Each cell label looks like
```
{ value: 1, cell: 2, z: 3 }
```
which says that the value 1 in frame 3 encodes cell 2. This is basically an association table (https://en.wikipedia.org/wiki/Associative_entity) for the many-to-many relationship between values and cells. Each value can encode many cells (so we can have overlapping cells at a pixel) and each cell can be encoded by many values. 

With the additional `z` field, each frame has a separate association table between cells and labels. By adding the `z` field, we can reassign the cells on each frame without changing the values written in the segmentation images, allowing for very snappy client side editing of the cell assignments. Note: in #331, `z` changes to `t`.

Undo/redo has major changes so edits to the segmentation and the cell labels are propagated. Communicating changes between dependent label types and grouping them for undo/redo is key functionality to generalize to new label types. Edits to one category of labels must propagate to the labels that reference them so the dependent labels can update themselves. This manifests in #331, as divisions contain cells, so updating the cell labels needs to be reflected in the division labels.

Editing the segmentation can change the cells by adding or deleting values, so we send the cell labels in cells.json along with the segmentation images together and get updates for both the segmentation image and the cell labels.


msschwartz21:
Can you add an example of what an overlap case would look like in this new format?

tddough98:
> Can you add an example of what an overlap case would look like in this new format?

Say cell 1 and cell 2 overlap where value 1 encodes cell 1, value 2 encodes cell 2, and value 3 encodes both cell 1 and cell 2. The cell labels would look like

```javascript
[
  { value: 1, cell: 1, z: 0 }, 
  { value: 2, cell: 2, z: 0 }, 
  { value: 3, cell: 1, z: 0 }, 
  { value: 3, cell: 2, z: 0 },
]
```

tddough98:
The loose ends in this PR are tied up in #331, so I'm marking it as ready to review so that the changes can be reviewed separately, even though there are still unfinished pieces in here.

tddough98:
> I can't quite track what happened to the apiMachine. Can you let me know where that code ended up in the reorg?

It's now segmentApiMachine and it's a child actor of segmentMachine.

> Is this the first time that the new undo/redo machines have been implemented? It looks great!

Not exactly, undo/redo machines for the UI state have been around since the react rework, and in #325 I added the labelHistoryMachines to track the labeled data state on the client too. This PR adds a labelHistoryMachine for the cell labels, and also generalizes what labels change for each edit. When the segmentation was the only thing changing, we could just make a stack of snapshots. Here, the cells may change without editing the segmentation, so we need a more general way to track what snapshots are associated with each edit.

PR #331, opened 2022-06-03
Replace lineage with divisions and add division editing

tddough98:
## Overview

This overhauls how divisions labels are encoded, changing to a event focused format `divisions.json`
```json
[{"parent": 1, "daughters": [2, 3], "t": 1}]
```
from the existing `lineage.json` format
```json
{
  "1": {
    "label": 1,
    "frames": [0],
    "frame_div": 1,
    "capped": true,
    "daughters": [2, 3],
    "parent": null,
  },
  "2": {
    "label": 2,
    "frames": [1],
    "frame_div": null,
    "capped": false,
    "daughters": [],
    "parent": 1,
  },
  "3": {
    "label": 3,
    "frames": [1],
    "frame_div": null,
    "capped": false,
    "daughters": [],
    "parent": 1,
  },
}
```

## Loading

The loader on the backend now looks for divisions.json in the input zip. For backwards compatibility, it supports loading lineage.json from .trk files and converts the lineage to divisions.json.

## Editing

Editing the divisions is handled by divisionsMachine. There is another separate `editDivisionMachine` to manage the UI state of editing divisions, like whether the user is being prompted to add a daughter to a division, which then sends events to divisionsMachine. Once the divisions change, we send a DIVISIONS event to the divisions event bus.

When the cell labels are edited, we forward these events (REPLACE, SWAP, or DELETE) so the divisions can also replace, swap, or delete the same cells present in division labels. "Replace with parent" and "create new cell" have been removed from the divisions tab as they are found in the cells tab instead and update the divisions.

Editing the segmentation also may change the cell labels. The segment edit API sends updates for the cell labels, but these are not yet propagated to the divisions. I'm still working on how to update divisions after segmentation edits. For now, as long as a parent or daughter cell is not deleted entirely using the segment API, the divisions stay valid.

Edits to the divisions are also recorded in a labelHistoryMachine, allowing division edits to be undone and redone. The divisionsMachine sends a SNAPSHOT event to its history machine to record what the divisions were before and after the edit.

The idbMachine listens to the divisions event bus for DIVISIONS event to persist the current divisions in IndexedDB.

# Other changes

Components that display divisions now use the hooks `useParentDivisions(cell)` and `useDaughterDivisions(cell)` to get the division events that the cell is involved in as a parent or daughter.

Swapping and replacing cell labels may make a cell a parent or daughter in multiple divisions. We now show all divisions where the selected cell appears.

## Minor code changes

- Rename Tracking to Divisions in UI text and component names
- Rename LabelControls to EditControls. This is the column just to the left of the canvas with the tools to edit the segmentation, cells, or divisions
- Rename `numFrames` to duration and `frame` to `t` or `time`
- Switch `z` to `t` in cell labels
- In label.py, `new_cell` and `new_value` are dynamically calculated instead of set at initialization so actions don't need to update them within the Edit class

## Update 6/9/2022

I've been squashing bugs with @msschwartz21 and tweaking functionality since putting label-dev.deepcell.org online. Here's a summary of new changes

- Display
  - Add color prop for OutlineCellCanvas
  - Show the daughter to be added on the canvas
  - Show outer outline of selected cell when it is the flood cell
  - Center each box in the cell timelines around markers on sliders
  - Prevent vertical white lines from appearing in timeline
- Performance
  - Run idbMachine in webworker
  - Switch from drawing canvases to drawing image bitmaps to improvement drawImage performance
- Usability
  - Handle segment edit API errors gracefully
  - Prevent swap and replace with self
  - Create two stages for add daughter prompt text
  - Clear second selected cell for flood, replace and swap when pressing Esc
  - Select a cell when using flood, replace, and swap if none selected
  - Add keybinds to use cell tools (V for select, Backspace for delete, S for swap, and R for replace)
  - Allow cell tools to be triggered by button presses, keybinds, clicks, or Enter
  - Switch opacity sliders from range slider to single slider
  - Click on cell timeline to jump between frames
- New functionality
  - Add divisions to export zip
  - Store spots in IndexedDB
  - Remove erased values before writing cells.json when completing an edit in label.py
  - Update divisions from EDITED_CELLS after editing segmentation
  - Enable flooding holes in cells by selecting the flood label as the select label
  - Code quality
  - Rename useSegment to useEditSegment
- Bug fixes
  - Fix creating too many WebGL contexts by sharing one canvas and GPU instance for all kernels that need to use premultipliedAlpha. Note: this may cause a memory leak as we can’t call kernel.destroy() without affecting the shared canvas
  - Fix loading divisions as None
  - Fix watershed to create a new cell
  - Enable raw based tools when in grayscale mode
  - Prevent re-registering edit actor with undo


tddough98:
> What does OuterOutlineCanvas do compared to OutlineCanvas?

OutlineCanvas and OutlineCellCanvas both make inner outlines, so the outline is fully inside the cell. OuterOutlineCanvas draws an outline just outside the cell. OutlineCanvas outline all cells, while OutlineCellCanvas outlines just one. OuterOutlineCanvas also outlines just one cell.

These components definitely could be abstracted better, as they only have slightly different behavior of what opacity to use, what color to use, whether to outline one cell or all cells, and whether to outline inside or outside the cell. I will definitely post an issue about unifying these components as I document the remaining and ongoing issues before I leave.

> What is the purpose of from-web-worker.ts?

This lets us spawn a WebWorker as a child of a state machine. We can pass message to the web worker by sending the spawned actors messages just like any other spawned actor instead of using the web worker [postMessage](https://developer.mozilla.org/en-US/docs/Web/API/Worker/postMessage) API.

PR #333, opened 2022-06-14
Update layout, instructions, exporting and remove visualizers

tddough98:
This contains some of the final changes needed before cutting the 0.4 release of DeepCell Label. It includes

- updating the in app drop down instructions
- move the display controls into a tab
- switching between tabs when using a keybind from another tab
- fixing bugs with too many WebGL contexts by having a shared GPU object with a single context

The changes are online at [label-dev.deepcell.org](https://label-dev.deepcell.org)


msschwartz21:
Some notes from my preliminary review of the new deployment:
- The selected cell always shows up as red in the selected cell icon
- I like how the instructions include actual demos of the controls, but can the formatting change so that the explanation shows up above or below the demo control?
Screen Shot 2022-06-14 at 8 17 44 PM
 - Layout on the Display instruction page is better with two consistent columns, but I still thinking it would be easier to read if everything were left aligned, or if the control column were narrower.
 - The fix suggestions for the different division warnings are great!

msschwartz21:
I think I found a bug with the timeline control. After using the timeline control demo in the instructions tab and switching to another instruction tab, the canvas stopped responding to the a/d keys. After I clicked on the primary timeline, it started responding to the key binds again.

msschwartz21:
Hitting "v" while on the cells tab, causes a return to the segment tab and activates the select button for segmentation. Otherwise the tab switching in response to key binds works well. E.g. selecting a cell while in segment and hitting the delete key leads to a switch to the cell tab and the cell is deleted.

tddough98:
All comments above are addressed and online at label-dev.deepcell.org (which now has these changes + changes from #334)

msschwartz21 (approved):

PR #334, opened 2022-06-16
Add forceLoadOutput URL parameter

tddough98:
To better support the Anolytics job workflow, we add a query string forceLoadOutput. When the query string is true, the page loads from the deepcell-label-output bucket. If the project is already in the local browser DB, we show a modal to confirm overwriting the local data. When the project has not been submitted, we show a missing project page with instructions to submit the project first.

For the backend to receive the query string to load from the output bucket, CloudFront needs to forward query strings, which it does not do by default. Setting the Origin request policy for the Elastic Beanstalk backend to AllViewer will forward all cookies, headers, and query strings. I will document this and all other cloudfront/s3/elastic beanstalk configuration in an upcoming deployment guide.

## Bug fixes
- prevent canvas from moving when changing tabs
- bug fix for undoing an action and then reloading before doing another action
- fix for loading PNGs with I or F mode (32 bit single channels)
- fix for setting dynamic range to 0 in grayscale mode


visualizer/src/Project/ForceLoadOutputModal.js
```diff
@@ -0,0 +1,58 @@
...
+  const id = new URLSearchParams(window.location.search).get('projectId');
+
+  return (
+    <Modal open={open} onClose={() => idb.send('USE_LOCAL_PROJECT')}>
+      <Box sx={style}>
+        <Typography variant='h6' component='h2'>
+          Confirm Force Reload from Output Bucket
+        </Typography>
+        <Typography sx={{ my: 2 }}>
+          Project <strong>{id}</strong> has been loaded previously, but you've opened a link to
```
msschwartz21, line 34: Screen Shot 2022-06-17 at 2 23 03 PM
The current deployment has extra text at the beginning of the message, but I'm not sure where it is coming from.

tddough98: Fixed in 7af53fe787aad573e0f681f15f74f68db6b9e758 and in current deployment on label-dev.deepcell.org

msschwartz21 (approved):
I used the forceLoadOutput flag a bunch over the weekend to re-export projects with the updated export code and it worked perfectly!

PR #368, opened 2022-08-23
Fix bugs related to cells not properly deleting

ykevu:
## What
Addressed the bug from Issue #365, in which after deleting a cell, one cannot write over the deleted cell's area when using exclusion mode. 

Additionally, added some snippets of code required for the tests and source code to run on Windows.

## Fix
The bug arose because the deleted cell was only removed from the cells matrix but its label persisted in the labels matrix. Added a new function to the Edit class called clean_labels() which deletes any labels that do not correspond to existing cells that is called every time an action is performed that adds masks (eg. brush, flood, etc.). Additionally, a more esoteric bug was found involving overlapped cells reappearing after deletion, which should be addressed by this change as well.

## Why
Users can now properly edit areas previously occupied by deleted cells, rendering the delete tool usable in exclusion mode.

### DIFF
--- a/backend/deepcell_label/label.py
+++ b/backend/deepcell_label/label.py
@@ -213,8 +213,7 @@ class Edit(object):
             (numpy array of shape (height, width), cells with updated values)
         """
         values = [cell['value'] for cell in cells]  # get list of values
-        cell_mask = np.in1d(labeled, values)
-        deleted_mask = np.logical_not(cell_mask).reshape(labeled.shape)
+        deleted_mask = np.isin(labeled, values, invert=True)
         labeled[deleted_mask] = 0   # delete any labels not in values
         return labeled
 
--- a/backend/deepcell_label/label.py
+++ b/backend/deepcell_label/label.py
@@ -166,6 +166,7 @@ class Edit(object):
         return mask
 
     def add_mask(self, mask, cell):
+        self.labels = self.clean_labels(self.labels, self.cells)
         if self.write_mode == 'overwrite':
             self.labels[mask] = self.get_value([cell])
         elif self.write_mode == 'exclude':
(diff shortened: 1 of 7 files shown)


tddough98:
The repeated `delete_temp` declarations can be factored out into `config.py` and imported from there. Also instead of `my_variable = True if condition else False`, use `my_variable = condition`.

In config.py,
```
DELETE_TEMP = platform.system() != 'Windows'
```
and then in each module that defines delete_temp locally, import `DELETE_TEMP` and replace `delete_temp` with `DELETE_TEMP`.

backend/deepcell_label/label.py
```diff
@@ -199,6 +200,24 @@ def clean_cell(self, cell):
...
+            labeled: numpy array of shape (height, width)
+            cells: list of cells labels like { "cell": 1, "value": 1, "t": 0}
+
+        Returns:
+            (numpy array of shape (height, width), cells with updated values)
+        """
+        values = [cell['value'] for cell in cells]  # get list of values
+        cell_mask = np.in1d(labeled, values)
+        deleted_mask = np.logical_not(cell_mask).reshape(labeled.shape)
+        labeled[deleted_mask] = 0  # delete any labels not in values
```
tddough98, line 218: `np.isin` would preserve the shape of labeled so you don't have to reshape after. Both `isin` and `in1d` have an invert parameter that can bypass the `logical_not`.

```
deleted_mask = np.isin(labeled, values, invert=True)
labeled[deleted_mask] = 0
```

backend/deepcell_label/loaders_test.py
```diff
@@ -89,12 +91,16 @@ def test_load_separate_npz():
 def test_load_image_tiff():
     """Load image from tiff file."""
     expected = np.zeros((1, 1, 1, 1))
-    with tempfile.NamedTemporaryFile() as images:
+    delete_temp = False if platform.system() == 'Windows' else True
```
msschwartz21, line 94: Import from config?

backend/deepcell_label/loaders_test.py
```diff
@@ -124,11 +130,15 @@ def test_load_image_and_segmentation_tiff():
 def test_load_image_png():
     """Load image array from png file."""
     expected = np.zeros((1, 1, 1, 1))
-    with tempfile.NamedTemporaryFile() as images:
+    delete_temp = False if platform.system() == 'Windows' else True
```
msschwartz21, line 133: Import from config?

msschwartz21 (changes requested):
Please address the other changes that Tom suggested and fix the linting error that is causing a test to fail. Then I would be happy to approve!

msschwartz21 (approved):

PR #372, opened 2022-08-27
Fix bug causing segmentations to be lost upon reload

ykevu:
## What
My guess is that after Tom fixed the bug related to adding c to cells, the "feature" field was renamed to "c", so the idbWebWorker was unable to find what segmentation feature was being edited, and segmentations were thus not being tracked.

## Why
Users can now reload without losing their segmentation changes (eg. brush adds/removes).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -161,5 +161,6 @@ npm-debug.log*
 yarn-debug.log*
 yarn-error.log*
 package-lock.json
+package.json
 
 .nyc_output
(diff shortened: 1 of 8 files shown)


msschwartz21:
Have you had a chance to test this version on label-dev? If you can confirm that users have tested it, then I am happy to approve.

ykevu:
> Have you had a chance to test this version on label-dev? If you can confirm that users have tested it, then I am happy to approve.

I deployed this version on label-dev, Elora and I both tested it, and changes are now correctly persisting.

msschwartz21 (approved):

Issue #367, opened 2022-08-19
Deleted cell reappears with label 0


elaubsch:
**Describe the bug**
Deleted cell reappears with label 0 after a second cell is deleted

**To Reproduce**

1. Ran Caliban prediction on deepcell.org and viewed results
2. Changed to CELLS tab and selected ALL mode
3. Selected a cell and deleted it
4. Selected another cell and deleted it
5. Old cell label reappeared
6. Hovered over the old cell with SELECT tool, saw label value is set to 0

**Expected behavior**
A deleted cell should not reappear.

**Screenshots**
![Screen Recording 2022-08-19 at 11 50 42 AM](https://user-images.githubusercontent.com/43769564/185688275-4053b983-44d0-48ab-9d07-dfe752260ba3.gif)

ykevu:
It seems like this is primarily a visual bug related to #365, so the fix to that issue should address this as well. With that change, cells do still reappear but making any further edits should clean up the improperly deleted cells, but I'll still look into why they reappear at all.

ykevu:
I was able to figure out what was causing this bug. The gpu library used to render the canvas seems to have a bug (?) where out-of-bounds indexing of an array gives unexpected values instead of undefined, so when deleting last cell(s) on a given frame, the kernel still thought the labels were there. By adding an if statement to check if the value the kernel is rendering is within the expected range of values, this bug should be fixed.

Issue #366, opened 2022-08-18
Not all segmented cells displayed in viewer


ngreenwald:
**Describe the bug**
After segmenting an image with Mesmer, when I clicked "View Results", only a subset of the segmented cells are displayed. 
[combined.tiff.zip](https://github.com/vanvalenlab/deepcell-label/files/9375539/combined.tiff.zip)

**Expected behavior**
All segmented cells should show up. Since the missing cells are all on the bottom, there's probably a maximum number of cells that can be displayed in a single image, and this value is hardcoded too low for dense tissue images. 

**Screenshots**
<img width="1084" alt="image" src="https://user-images.githubusercontent.com/13770365/185442333-bd0fbe39-b696-4d88-a479-69baf8362f0d.png">

**Additional context**
I attached the image I used to make the prediction to the issue. Here's the URL: https://label.deepcell.org/project?projectId=aJS695DpYT7n&download=true

ykevu:
Hi Noah, I was able to find the parameter that results in only 999 of the cells in that file rendering on the canvas (the rendering kernel has a maximum loop iteration count of 1000). My background is not really in biology, so what would you consider a better upper limit? Do you think it reasonable to assume that an image would never need to render over 5000 cells?

ngreenwald:
Yes, I would say 5000 is a good conservative estimate, at least for as long as the max image size remains 2048. Once that gets larger then the number of cells could increase as well

PR #376, opened 2022-09-08
Fix canvas rendering bugs caused by issues with gpu rendering kernel

ykevu:
## What
Fixes canvas rendering bugs, namely #366 and #367, which were both caused by issues in the gpu rendering kernel. Specifically, a check is now called to ensure that a value is contained within the bounds of expected values before trying to render the label associated with that value (out of bounds indexing was resulting not in errors but the cells reappearing upon deletion). Additionally, the max number of cells displayed has been increased to 5000.

These changes were tested by Changhua and me and appear to be working properly.

## Why
Users can now render up to 5000 cells at once on a canvas and will not experience bugs when deleting the highest valued cells in a given frame.

### DIFF
--- a/backend/deepcell_label/blueprints.py
+++ b/backend/deepcell_label/blueprints.py
@@ -75,7 +75,9 @@ def create_project():
         )
     labels_url = request.form['labels'] if 'labels' in request.form else None
     axes = request.form['axes'] if 'axes' in request.form else None
-    with tempfile.NamedTemporaryFile(delete=DELETE_TEMP) as image_file, tempfile.NamedTemporaryFile(delete=DELETE_TEMP) as label_file:
+    with tempfile.NamedTemporaryFile(
+        delete=DELETE_TEMP
+    ) as image_file, tempfile.NamedTemporaryFile(delete=DELETE_TEMP) as label_file:
         if images_url is not None:
             image_response = requests.get(images_url)
             if image_response.status_code != 200:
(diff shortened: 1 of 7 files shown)


backend/deepcell_label/blueprints.py
```diff
@@ -75,7 +75,9 @@ def create_project():
         )
     labels_url = request.form['labels'] if 'labels' in request.form else None
     axes = request.form['axes'] if 'axes' in request.form else None
-    with tempfile.NamedTemporaryFile(delete=DELETE_TEMP) as image_file, tempfile.NamedTemporaryFile(delete=DELETE_TEMP) as label_file:
+    with tempfile.NamedTemporaryFile(
+        delete=DELETE_TEMP
+    ) as image_file, tempfile.NamedTemporaryFile(delete=DELETE_TEMP) as label_file:
```
msschwartz21, line 80: I thought the operating system changes were introduced in another PR. Is this commit left over from old branch history or were there some changes that didn't make it into the last PR?

frontend/src/Project/Canvas/LabeledCanvas.js
```diff
@@ -71,6 +73,7 @@ export const LabeledCanvas = ({ setBitmaps }) => {
         output: [width, height],
         graphical: true,
         dynamicArguments: true,
+        loopMaxIterations: 5000,
```
msschwartz21, line 76: Can you add a comment mentioning that this controls the max number of cells on the canvas?

msschwartz21 (commented):
Two quick questions, but otherwise this looks good to go.

backend/deepcell_label/blueprints.py
ykevu: This is left over from old history; I think my branch was not up to date, sorry

msschwartz21 (approved):

PR #378, opened 2022-09-21
Fix bugs breaking brush tool and unintended undo behavior

### LINKED ISSUE #377: Brush tool stops working when switching to another tab and back + on undo
**Describe the bug**
After performing certain actions, switching back to the brush tool with either the 'b' hotkey or by clicking the brush tool will not allow you to perform brush strokes (mouse just stalls at area you click as if still in select mode).

**To Reproduce**
Unclear exactly what conditions cause this to occur, but in general it seems that if you switch from brush to delete and back, it won't work unless you reselect.

ykevu:
## Significant Changes (Fix #377)
- Fix bug where brush tool broke when switching to another tab and back.
    - EditMachine.js now sends ENTER_TAB to EditSegmentMachine.js when the SEGMENT tab is selected. EditSegmentMachine will 
 accordingly send SET_PAN_ON_DRAG depending on the state it is in so as not to potentially break the state.
- Fix bug where undoing from a pan tool to a non-pan tool (brush) would break the brush.
    - panOnDrag state is now stored in the undo history for CanvasMachine.js, and upon restoring, a CHECK_DRAG event is sent to restore the appropriate state.

## Small changes
- Fix bug where using CTRL+Z to undo would always switch to the DISPLAY tab.
    - An event listener was listening for Z presses to switch to DISPLAY tab, so Z was removed from this and a mousetrap binding was instead created to not trigger on combination hotkeys
- Drop uploading now actually uses the user-selected axes
- .zip added to list of acceptable file formats (application/zip was not working on my Windows machine)
- Added hotkey tooltip on select (V)


frontend/src/Project/EditControls/SegmentControls/ToolButtons/SelectButton.js
```diff
@@ -9,7 +9,11 @@ function SelectButton(props) {
 
   const onClick = useCallback(() => segment.send({ type: 'SET_TOOL', tool: 'select' }), [segment]);
 
-  const tooltipText = <span>Click to select a label</span>;
+  const tooltipText = (
+    <span>
+      Click to select a label <kbd>V</kbd>
+    </span>
+  );
```
msschwartz21, line 16: How is the `V` key bind different from hitting esc to default to select mode?

msschwartz21 (approved):

frontend/src/Project/EditControls/SegmentControls/ToolButtons/SelectButton.js
ykevu: I believe the purpose of hitting `esc` is to "exit" from the current selection (eg. if you have cell 10 selected, you will reset to "cell 0" meaning nothing is selected), and the currently used tool will also persist. The purpose of hitting `v` is to actually change the current tool to select mode (but the current selection is still maintained). 

In fact, because hitting `esc` does not change the tool, I think I discovered a bug where hitting `esc` while using the brush tool allows you to brush cells with value 0, which I'll raise an issue for.

Issue #373, opened 2022-08-28
Changes to file export to match lab conventions


msschwartz21:
The current file export makes some choices that do not match lab conventions for how we store our data. In order to simplify the process of retrieving files from DCL jobs, I would like to slightly change the export to match our conventions.

1. Change shape of the axes. Current export axis order is `CZYX` which should be changed to `ZYXC`.
	https://github.com/vanvalenlab/deepcell-label/blob/f374ffadfee40d12eb931e11dc14075c6f6eb152/backend/deepcell_label/export.py#L91
	https://github.com/vanvalenlab/deepcell-label/blob/f374ffadfee40d12eb931e11dc14075c6f6eb152/backend/deepcell_label/export.py#L103
This should be a straightforward change but we should test loading some of the exported images in Fiji and a notebook to confirm that the export is reasonable.
2. Raw data (X) should not be permanently rescaled to 0-255. Currently the data loader saves a new ome.tiff after loading from the original file format.
https://github.com/vanvalenlab/deepcell-label/blob/f374ffadfee40d12eb931e11dc14075c6f6eb152/backend/deepcell_label/loaders.py#L82-L84
Ultimately we do not want to change the dynamic range and dtype of our raw data. I would like to change these lines to eliminate the rescaling. However it may introduce downstream issues with displaying the data, so we can discuss further as those potential issues are discovered.

msschwartz21:
I spent some time looking in the ome.tiff and ome.xml specifications to try to find an explicit constraint on dimension order. Unfortunately the documentation included below really doesn't help. It implies that dimension order should always start with `XY` even though Tom's original specification of `CZYX` clearly seemed to work.

http://www.openmicroscopy.org/Schemas/Documentation/Generated/OME-2016-06/ome.html

<img width="958" alt="Screen Shot 2022-09-02 at 2 15 32 PM" src="https://user-images.githubusercontent.com/20373588/188213916-6acee6ed-5839-4de7-b3ce-685f5967a1cf.png">

ykevu:
I saw this as well, but every library I tried has enforced that 'YX' be the last two dimensions, so I think there's some sort of reversal at play. The documentation is usually non-existent or unclear as well. The 'axes' field that Tom used for tifffile has the following description:

![image](https://user-images.githubusercontent.com/55016297/188215703-c9062de9-b444-439a-a2e5-8462e7ef1f69.png)

PR #383, opened 2022-09-28
Change raw rescaling to happen on frontend + export original raw unscaled

ykevu:
## PR Addressing #373
* Non-png raw inputs are no longer rescaled at all on the backend in loaders.py (.png files are still rescaled to uint8)
* Instead, the frontend performs the rescaling on the raw array, and creates a second array "rawOriginal" which is passed to the arrayMachine. 
    * This is a "TypedArray" which can support Uint8, Int8, Uint16, Int16, Uint32, Int32, Float32, and Float64. This will require further testing (by @steveyu323?) for a variety of dtypes since I only had access to uint8 and float64 files
* On export, rawOriginal is zipped instead of the rescaled raw array, so download or submission will give the user the original raw unscaled tiff.
### Testing
* So far, I have tested uploading a file from Changhua that is float64, and the rendering on the site was fine. Upon downloading, the X.ome.tiff file inside retained float64 and the unscaled values. After reuploading this downloaded file to DCL, it looked fine as well. Functionality of all the editing and such still worked throughout all of this.
### Potential Issues / Enhancements
* The loading times do not seem to have suffered very significantly, but will look into using gpu.js to speed up the reshaping operation (which currently has 4 nested for loops).

### Small changes
* REPLACE tool was renamed to COMBINE on its button, tooltip, and instructions. The hotkey and code referencing the tool is unchanged.

### DIFF
--- a/frontend/src/Project/service/loadMachine.ts
+++ b/frontend/src/Project/service/loadMachine.ts
@@ -72,10 +72,11 @@ async function splitArrays(files: Files) {
   const labeledFile = files['y.ome.tiff'] as OmeTiff;
   const raw = await getRawRasters(rawFile.data[0]);
   const labeled = await getLabelRasters(labeledFile.data[0]);
-  return { raw, labeled };
+  const rawOriginal = await getRawOriginalRasters(rawFile.data[0]);
+  return { raw, labeled, rawOriginal };
 }
 
-async function getRawRasters(source: TiffPixelSource) {
+async function getRawOriginalRasters(source: TiffPixelSource) {
   const { labels, shape } = source;
   const c = shape[labels.indexOf('c')];
   const z = shape[labels.indexOf('z')];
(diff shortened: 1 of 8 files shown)


frontend/src/Project/service/loadMachine.ts
```diff
@@ -111,15 +144,47 @@ async function getLabelRasters(source: TiffPixelSource) {
   return channels;
 }
 
-type Raster = { data: Uint8Array | Int32Array; width: number; height: number };
+function reshapeRaw(channels: TypedArray[][][], min: number, max: number) {
+  // Normalize each pixel to 0-255 for rendering
+  for (let c = 0; c < channels.length; c++) {
+    for (let z = 0; z < channels[c].length; z++) {
+      for (let y = 0; y < channels[c][z].length; y++) {
+        for (let x = 0; x < channels[c][z][y].length; x++) {
```
msschwartz21, line 152: Is it really necessary to iterate over x and y too? Can you not apply the rescaling to an entire frame?

msschwartz21 (changes requested):
Can you look into why the coverage decreased so dramatically? I don't mind small drops but a 64% drop seems to indicate a problem in how the tests are running.

@steveyu323 would you mind testing some different file uploads before we merge this in?

steveyu323:
> Can you look into why the coverage decreased so dramatically? I don't mind small drops but a 64% drop seems to indicate a problem in how the tests are running.
> 
> @steveyu323 would you mind testing some different file uploads before we merge this in?

Sure thing!

frontend/src/Project/service/loadMachine.ts
ykevu: You can apply rescaling to an entire frame using forEach but it seems like these sorts of callback functions are actually slower than regular for loops. I will do some benchmarking and try to use gpu.js to speed it up

ykevu: > Can you look into why the coverage decreased so dramatically? I don't mind small drops but a 64% drop seems to indicate a problem in how the tests are running.

I reran the backend tests and the coverage appears normal now; I'm not sure what happened on the initial run.

msschwartz21 (approved):

PR #402, opened 2023-02-01
Implement centroid calculations and channel expression unit testing

ykevu:
## What
* Adds ability to calculate centroids of each cell in order to spatially visualize them on the plotly chart and graph them
* Adds unit testing for both the new centroid calculations and for the original mean and total calculations in channelExpressionMachine.js


rossbar (approved):
Thanks @ykevu !

Issue #413, opened 2023-02-03
Allow training and calculations over all time frames


ykevu:
Currently, when performing in-browser training or a raw calculation, only the labeled frame at the current time is considered, however, we want to be able to consider data across all frames (eg. if working with a tiled image). Additionally, this leads to the training machine breaking when data exists in other frames. The ability for a user to make this distinction and obtain desired results should be implemented.

ykevu:
Implemented with merge of #424

PR #424, opened 2023-02-08
Allow calculations and training across all frames

ykevu:
## What
Addresses #413
* Implemented functions to calculate and visualize mean/total embeddings across all frames
* Implemented UI controls that allow users to specify whether they want to calculate/visualize/train/predict on the current frame or across all frames
* Added documentation to trainingMachine functions since those functions are a bit of a mess
* Added unit tests for new cross-frame calculation functions


rossbar (approved):
LGTM, thanks @ykevu !

PR #441, opened 2023-04-11
Add button for deleting channels, remove original display controls

ykevu:
It can be cumbersome to remove channels due to the remove button being accessible only through a popup menu, so I added a trash icon button so users can quickly delete channels they don't want to see.

Additionally, the original display controls on the left are now removed in favor of purely using the top display control panel.


rossbar (approved):
It'd be nice to add a test, but LGTM!

ykevu:
> It'd be nice to add a test, but LGTM!

@rossbar Modified the test to accommodate this change and also ended up removing all the old display controls and logic.

PR #449, opened 2023-04-25
Marker panel visualization and editing modal

ykevu:
Adds a popup component that allows the user to visualize the marker panel and make edits. There are still a number of features that would be useful to implement, but I wanted to get this in soon and move on to more important things, since, though it looks pretty, in practicality this feature shouldn't really be used much (hopefully, if our project creation scripts work as intended, not at all) by an annotator such as Martin, other than as a non-interactive reference.

Specifically, the following **is implemented** by these changes:
* The user can click an icon that pops out the modal with a table visualizing the marker panel, with colors/lit up chips that denote whether the cell type or channel is matched with the current image
* The user can double click cells to make edits to the cell types or channels in the marker panel
* Channels now must be case-sensitive, exact matches to be detected (this can lead to inconveniences for sure, so am still thinking about how best to approach this, but the projects for Martin should have properly mapped channel names and such anyway)

The following is **not** implemented that would theoretically be useful in general cases:
* Persistence of edits, whether through import/export or IndexedDB
* Tracking of edits with UNDO/REDO (I'm not even sure this would be desirable)
* Ability to add new entries to the marker panel (ie. cell type that we have not considered)
* A visual list of the image's channels to reference when constructing a marker panel


rossbar (approved):
Thanks @ykevu - I agree this is an important feature and we should get it in ASAP and worry about additional features/tweaks in followup PRs

PR #450, opened 2023-05-05
Add toggle to NavBar to switch between light and dark mode

ykevu:
Adds an icon button to the navbar in the top right to toggle between light and dark mode themes. It turned out that two lines of default dark mode from MUI and adding the cssbaseline component made this essentially good enough out of the box.

Lower priority TODOs:
- The plotly components are still blinding white in dark mode, but they would definitely be a pain to sync up
- Cypress test


rossbar (approved):
Great! Thanks @ykevu !

PR #461, opened 2023-05-08
Celltype UI toolbar with new toggles

ykevu:
## What
* The cell type controls panel on the left now has a toolbar instead of the giant ADD CELLS button. It includes:
1. Add Cells button (same as before, but as an icon now)
2. Marker Panel open button (same as before, but as part of the toolbar now)
3. Cell type writing mode toggle (Defaults to "overwrite," so that adding Cell 1 which is labeled Type A to Type B removes it from Type A). This addresses #453.
4. Cell type hovering toggle (Defaults to off, but if on, you can hover over cells to view what cell type(s) it is in). This addresses #452.
* There are also tooltips for each button, but a TODO is to add keybindings for them in some form


rossbar (approved):
Thanks @ykevu !

PR #465, opened 2023-05-22
Fix memory leak caused by continuously returning OutlineCellCanvas

### LINKED ISSUE #460: Memory leak
It appears that some sort of memory leak is present, as performing cell type related actions can quickly increase the memory usage until the app crashes. Some potential culprits are the cell type color mapping array recalculation and undo/redo, but profiling will have to be done

ykevu:
## What
In AddCellTypeCanvas.js and AddDaughterCanvas.js, we have:
```python
 if (!cell) {
    return null;
  }
```
I don't have a deep enough understanding of React to know exactly why is the case under the hood, but because of this code, the OutlineCellCanvas component is regenerated without disposing of its memory every time a cell goes from selected to not selected in the context of add mode (which equates to adding a cell to a cell type -> going from returning the component to returning null, alternating). This is especially bad because OutlineCellCanvas calls useCellMatrix(), which is a hook that, if needed, reallocates an approximately N x N matrix where N is the highest ID cell in your current frame. 

Ultimately, this means that for the Keren dataset approximately 120 mb are reallocated everytime you add a cell in the final frame. I'm fairly confident that the issue is gone when this piece of code is removed though (whew).

Closes #460

### DIFF
--- a/frontend/src/Project/Canvas/ToolCanvas/AddCellTypeCanvas.js
+++ b/frontend/src/Project/Canvas/ToolCanvas/AddCellTypeCanvas.js
@@ -10,10 +10,6 @@ function AddCellTypeCanvas({ setBitmaps }) {
   const editCellTypes = useEditCellTypes();
   const cell = useSelector(editCellTypes, (state) => state.context.cell);
 
-  if (!cell) {
-    return null;
-  }
-
   return <OutlineCellCanvas setBitmaps={setBitmaps} cell={cell} color={white} />;
 }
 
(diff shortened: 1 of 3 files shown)


ykevu:
It seems like this article does a good job of explaining why returning null is a bad idea in React: https://medium.com/@davidkelley87/stop-using-return-null-in-react-a2ebf08fc9cd

We should be on the lookout for any other components where we might have done this!

rossbar (approved):
Awesome, thanks @ykevu for tracking this down!

PR #469, opened 2023-05-25
Feature: drag celltype cards for reordering

ykevu:
## What
* Enables the user to drag cell type cards to reorder them--this is purely a UI feature and does not affect the underlying cell types data at all
* Introduces this with the react-beautiful-dnd library
### Notes
* The opacity slider is now wrapped in a bit of a gross div as a workaround for the drag behavior not recognizing MUI sliders and thus not allowing you to drag the slider

### DIFF
--- a/backend/deepcell_label/export.py
+++ b/backend/deepcell_label/export.py
@@ -19,6 +19,7 @@ class Export:
         self.load_dimensions()
         self.load_labeled()
         self.load_raw()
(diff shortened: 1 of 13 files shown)


ykevu:
@rossbar I'm going to go ahead and merge these last PRs in, since the CI issues are fixed and these are already deployed anyway!

PR #471, opened 2023-06-03
Refactor idbWebWorker to separate indexedDB storage into multiple obj…

ykevu:
…ect stores

## What
* We noticed that any time some sort of edit is made (including adding a single cell to a cell type), there would be a large amount of disk usage and waiting before the IndexedDB was done updating. 
* This was because we were using one object to store all persisted context in the IndexedDB, so even adding one element to the cellTypes json, for example, entailed rewriting the entire raw image, labeled mask, etc. to IndexedDB, which was very undesireable especially with larger images
* By separating into separate object stores instead, we can update cellTypes.json, for example, while leaving the raw arrays untouched

### DIFF
--- a/frontend/src/Project/service/idbWebWorker.js
+++ b/frontend/src/Project/service/idbWebWorker.js
@@ -18,14 +18,14 @@ const idbMachine = createMachine(
     context: {
       db: null,
       projectId: null,
-      project: {
-        raw: null, // Uint8Array[][][]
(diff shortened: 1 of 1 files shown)


rossbar (approved):
Thanks for diving in @ykevu , this seems like a sensible approach to better manage resource utilization. Please feel free to modify/merge as you see fit!

Issue #470, opened 2023-05-25
Figure out the issue with using useSelector() in some cases


ykevu:
**Describe the bug**
This was most notable in leading to janky workarounds for both the cell type toggles and the cell type orders (from drag and drop). Essentially, React is not updating when we make changes to state context sometimes. 

With the cell type orders from #457, for example, simply assigning the cell types context to its reordered state after a drag and drop event does not lead to the React component re-rendering from the change in state. Since re-renders occur for anything that is being recorded into UNDO history, it is probably worth looking into why this discrepancy is occurring and fixing this. The implications being that the order of the cell types can be saved into the machine state (right now, the order will reset upon switching tabs, for example, since the order state is only tracked within the accordion list component)

ykevu:
The cell type order issue was actually unrlated, but this is still important for the toggle logic

ykevu:
See #473

Issue #427, opened 2023-02-24
Toggle logic breaks after undoing cell type deletion, for example


ykevu:
**Describe the bug**
Deleting a cell type and then undoing the deletion can break the toggle checkboxes

ykevu:
Deleting cell types in general can seem to trigger this behavior

ykevu:
Doing toggles in add mode can also break this, since it updates the toggle state in the components but not in the state machine

ykevu:
Some of this will be fixed with #473 , but the UNDO issue and the add mode issue will still need to be fixed.

PR #473, opened 2023-06-08
Move toggle logic back to state machine, fix various bugs

ykevu:
## What
* The array that controls what cell types are toggled on or off is moved back to the state machine, as it should be--I found that the fix to #470 was because we have to reassign arrays to new arrays for React to tell that a change was made (the Javascript equivalent of having to assign a variable to list.copy() instead of assigning to list)
* I also fixed bugs relating to how the array is initialized and changed when cell types are turned on, turned off, deleted, added, etc. in various combinations
* **TODO**: The array will still bug out if you, say delete a cell type and then undo that deletion (see #427). This can be fixed by making cellTypesMachine register the UI and have the toggle logic (and opacities should fall under this too) tracked with UNDO/REDO; however, this is a bit of a pain and there are more prioritized things that need to be done, so that will be for a future PR and I'll reopen a issue

### DIFF
--- a/frontend/src/Project/EditControls/CellTypeControls/CellTypeControls.js
+++ b/frontend/src/Project/EditControls/CellTypeControls/CellTypeControls.js
@@ -2,8 +2,6 @@
 // the list of cell types, and an editing prompt when adding cells
 
 import { Box, FormLabel } from '@mui/material';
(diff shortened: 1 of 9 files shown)


ykevu:
@rossbar I'm going to start merging these in!

PR #483, opened 2023-06-26
Add API endpoint for creating projects from zip file

ykevu:
Adds an API endpoint for creating projects with a zipfile opened in memory. This should make it easy to programmatically create DCL projects by simply sending a POST request to /api/project/zip with a field for 'zip' @rossbar ! 

I think this could probably be consolidated with the 'dropped' endpoint as well, but this is the simple and fast solution for now.

### DIFF
--- a/backend/deepcell_label/blueprints.py
+++ b/backend/deepcell_label/blueprints.py
@@ -114,6 +114,38 @@ def create_project():
     )
     return jsonify(project.project)
 
+@bp.route('/api/project/zip', methods=['POST'])
+def create_project_from_zip():
+    """
+    Create a new Project from a local zipfile variable.
+    """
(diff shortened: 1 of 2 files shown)


ykevu:
@rossbar I realized that we can just use the dropped endpoint already; I guess I did something wrong when I tried it in the past. For instance, if we have a zip in local disk, we can just do:
```Python
import requests

url = '____/api/project/dropped'
axes = 'CZYX'
# Replace with the path to the local zipfile
zipfile_path = '___.zip'

with open(zipfile_path, 'rb') as f:
    files = {'images': f}
    data = {'axes': axes}
    response = requests.post(url, files=files, data=data)
    print(response.text)
```

Issue #484, opened 2023-06-28
Cell expression bar graph automatically makes some channels disappear if there are enough of them


ykevu:
Will need to fiddle with the Plotly component to enable dynamic resizing depending on channel number, perhaps

ykevu:
I've seen this issue arise again in various cases, it might be a little tricky to completely solve this since the Plotly settings are so finicky

PR #485, opened 2023-06-28
Adjust height and margin of cell expression plot

ykevu:
## What
* Adjusts the height and margin of cell expression plot to account for larger number of channels and variety of channel name lengths as referenced in #484 . This is difficult to test since we don't have a easy way of changing the channel names list so this may have issues again in the future, but the plot looked good for a variety of lists that I looked at (Hickey, Liu, 2d example)


rossbar (approved):
LGTM, let's give it a try!

PR #524, opened 2023-10-04
added segment all button and mask rendering

mimithecoconut:
This adds a Segment-SAM tab with an Segment All button and the corresponding mask rendering connections. Clicking Segment All should segment according to the given mask. Also, modifies processing for npy file which had a bug before. Disables immediate segmentation loading but segmentation can be shown by using Segment All button.


ykevu:
Out of curiosity, is there a specific reason the SAM button needs to be on its own tab?

mimithecoconut:
> Out of curiosity, is there a specific reason the SAM button needs to be on its own tab?

Because there are going to be additional SAM actions to add in the tab and also Dave said to put it in a separate tab.

rossbar (commented):
Unfortunately it looks like this PR has the "windows vs unix line-endings" issue that prevents it from really being reviewable. I'll take a stab at that when I get a chance but @mimithecoconut it might be worth reviewing your editor configuration again! FWIW the thing that we really need is that diffs *only* reflect the actual changes rather than a re-writing of every file that was opened!

mimithecoconut:
> Unfortunately it looks like this PR has the "windows vs unix line-endings" issue that prevents it from really being reviewable. I'll take a stab at that when I get a chance but @mimithecoconut it might be worth reviewing your editor configuration again! FWIW the thing that we really need is that diffs _only_ reflect the actual changes rather than a re-writing of every file that was opened!

Yeah I realized that two of the files got completely all modified due to that. Trying to fix it. Also I saw issue with test and trying to figure out why the backend tests don't pass.

